PR #2, opened 2020-06-24
Setup armando folder with initial files

aroquev00:
Created my folder and included all the initial files from the STEP repo, modified my README.md to include my name.


brettapeters:
Can you just add the files under portfolio/ instead of the entire STEP repo including walkthroughs etc.? So you'll have something like this:
```
step191-2020/
+-- aroquev/
  +-- src/
  +-- README.md
  +-- pom.xml
```

aroquev00:
@brettapeters Sure! Done.

PR #6, opened 2020-06-24
Toggle function and corrections

pamelalozano16:
I did the conversion on the toggle function to js with event listener and class toggle. Also, when I compiled the web page it generated another target folder so I created a git ignore. And I also tried squashing the commits, did I do it right?


pamelalozano16:
@brettapeters Can you check the changes please? I wanted to ask for help here before I make the request in the other repo.

pamelalozano/src/main/webapp/index.html
```diff
@@ -88,7 +87,7 @@ <h4>Press &amp Charge</h4>
     </div>
     </div>
     
-<script src="https://code.jquery.com/jquery-3.2.1.slim.min.js" integrity="sha384-KJ3o2DKtIkvYIK3UENzmM7KCkRr/rE9/Qpg6aAZGJwFDMVNA/GpGFF93hXpG5KkN" crossorigin="anonymous"></script>
+<script src="script.js"></script>
```
brettapeters, line 90: This should be the last script before \</body> in case you need to reference anything from popper or bootstrap. I don't think it will make a difference here, but it's a good practice.

pamelalozano/src/main/webapp/script.js
```diff
@@ -27,14 +27,9 @@ function addRandomGreeting() {
...
-
-    if($("#c-"+id)[0].style.display!="block"){
-        $("#c-"+id)[0].style.display="block"
-        location.href = "#c-"+id;
-    } else{
-          $("#c-"+id)[0].style.display="none"
-    };
-}
\ No newline at end of file
+document.querySelectorAll('.project').forEach(item => {
```
brettapeters, line 30: This is definitely an improvement over the last version!

There is still room for improvement by using event delegation and bubbling i.e. put a single event listener on the projects parent container element (div with ID "#projects") and let the click events on its children bubble up to its event listener. In that event handler callback you'll have to figure out which child element was actually clicked by looking at `event.target` and possibly some other properties of event.target to figure out if you need to do anything.

Very good examples here: https://davidwalsh.name/event-delegate

pamelalozano16: Solved!

pamelalozano/src/main/webapp/index.html
pamelalozano16: Done! Thanks

pamelalozano/src/main/webapp/script.js
```diff
@@ -27,14 +27,7 @@ function addRandomGreeting() {
...
-    if($("#c-"+id)[0].style.display!="block"){
-        $("#c-"+id)[0].style.display="block"
-        location.href = "#c-"+id;
-    } else{
-          $("#c-"+id)[0].style.display="none"
-    };
-}
\ No newline at end of file
+document.getElementById("projects").addEventListener("click", function (event) {
+    event.target.closest('.project').classList.toggle("open");
```
brettapeters, line 31: Since clicks on a ".project-desc" or the parent "projects" element will also be handled by this callback, you might want to also make sure the event.target is a ".navbar".

For example, if you click on "div#projects", it will not find a closest ".project" and return null. You'd get an error like "Cannot find property classList of null" or something.

pamelalozano16: Oh you are right! It's done.

pamelalozano/src/main/webapp/index.html
```diff
@@ -125,6 +128,76 @@ <h3>A little taste of my life:</h3>
 </div>
 </div>
     
+<div class="container" id="skills">
```
pamelalozano16, line 131: I though I had a commit of this section, my mistake. 
I added the skills section from a template.

PR #1, opened 2020-06-23
Add fire

brettapeters:

PR #26, opened 2020-06-29
Cars Feature

aroquev00:
Added a section that at the click of a button shows a random car of a list of personal favorites and some info about it. Every time the button is clicked a new car is shown.

## PR Checklist

- [x] PR has a descriptive title
- [x] PR contains closely related changes only (should not have many unrelated changes)
- [x] STEP pod members have been added as reviewers

## PR Type

- [ ] Bugfix
- [x] Feature

Nonexistent feature

Issue Number: N/A

## What is the new behavior?
The click of the new button  shows an image of a car and 4 lines of details.

image

## Other information
Closes https://github.com/aroquev00/step191-2020/issues/1


brettapeters:
I think this branch was branched from your fork's messed up master before the fork's master was cleaned up to be in sync with the upstream. That's why there are many older commits in the branch that do belong to any branch on googleinterns:master.

I think we should just leave it like this and it'll get sorted out by squashing. Delete this branch after it's merged and start your future branches from your clean `master` branch.

aroquev00:
👍

brettapeters:
Can you use 2 spaces instead of tabs? I think that makes code more readable in GitHub.

aroquev00:
Hmmm, I set it up so that it indents with 2 spaces...

aroquev00:
image
does this mean It's indenting with a tab thats equivalent to 2 spaces? Or should I anyway change it to indenting with 2 spaces. I tried changing it to 2 spaces and stays the same... But maybe it just works in VSCode and github understands the encoding differently?

brettapeters:
In VSCode you'd have to change it to "Indent using spaces" and use "Spaces: 2". Then hitting the "tab" key will insert 2 spaces. Might look the same in VSCode, but GitHub will show the tab character as much more whitespace.

aroquev/src/main/webapp/script.js
```diff
@@ -62,25 +62,108 @@ function showPreviousSlide() {
...
+			country: "Germany",
+			funFact: "Grandpa's car <3, learned to drive in it.",
+			path: "images/vw.jpg"
+		}
+	];
+
+	// Pick a random car.
+	const car = cars[Math.floor(Math.random() * cars.length)];
+
+	// Add it to the page.
```
brettapeters, line 154: You should construct the whole DOM structure for a car in JS and attach it to the parent car view container rather than swapping out the text of several elements. This way the JS is not coupled with the HTML. If you wanted to change/add an attribute of car you would only have to do it in one place.

For example

```js
buildCarView(car) {
  const carView = document.createElement('div');

  const image = document.createElement('img');
  image.src = car.path;
  image.classList.add('car-image');

  const name = document.createElement('div');
  name.textContent = `Name: ${car.name}`;

  // More elements...

  carView.appendChild(image);
  // append stuff to carView

  return carView;
}
```

Inside of `getRandomCar()`
```js
const carContainer = document.getElementById('car-container');
const carView = buildCarView(car);
// replace contents of carContainer with carView
```

brettapeters: FWIW it is also waaaay faster for the browser to construct DOM nodes in JS rather than perform updates bit-by-bit.

seanjohnite:
Link to deployed version? : )

aroquev00:
https://aroquev-step-2020.uc.r.appspot.com/ :D

aroquev/src/main/webapp/script.js
```diff
@@ -62,25 +62,108 @@ function showPreviousSlide() {
...
+	dots[slideIndex].classList.add("active");
 }
 
 initSlides();
+
+/**
+ * Adds a random car and its data to the page.
+ */
+function getRandomCar() {
+	const cars = [
```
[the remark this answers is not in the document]
aroquev00: True, done.

aroquev/src/main/webapp/script.js
aroquev00: Wow, nice! Done.

PR #35, opened 2020-07-01
PR for walkthrough #3 of Week 3

aroquev00:
Modify Button that when clicked triggers a fetch for the Java servlet, which returns facts of cities in JSON, and are parsed and added to the DOM to be displayed in the HTML.

image

### DIFF
--- a/aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
+++ b/aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
@@ -27,6 +27,6 @@ public class DataServlet extends HttpServlet {
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     response.setContentType("text/html;");
-    response.getWriter().println("<h1>Hello world!</h1>");
+    response.getWriter().println("<h1>Hello Armando!</h1>");
   }
 }
--- a/aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
+++ b/aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
@@ -17,7 +17,6 @@ package com.google.sps.servlets;
 import com.google.gson.Gson;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.List;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
@@ -27,14 +26,15 @@ import javax.servlet.http.HttpServletResponse;
 @WebServlet("/data")
 public class DataServlet extends HttpServlet {
 
-  private ArrayList<String> comments;
+  private static final ArrayList<String> comments = new ArrayList<String>();
+  comments.add("I would like to live in NYC.");
+  comments.add("Berlin looks like a nice place to have fun.");
+  comments.add("I gotta go visit Japan one day!");
 
   @Override
   public void init() {
-    comments = new ArrayList<>();
-    comments.add("I would like to live in NYC.");
-    comments.add("Berlin looks like a nice place to have fun.");
-    comments.add("I gotta go visit Japan one day!");
+    
+    
   }
 
   @Override
@@ -50,7 +50,7 @@ public class DataServlet extends HttpServlet {
   /**
    * Convert to JSON using Gson
    */
-  private String convertToJson(ArrayList comments) {
+  private String convertToJson(ArrayList<String> comments) {
     Gson gson = new Gson();
     String json = gson.toJson(comments);
     return json;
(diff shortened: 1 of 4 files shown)


aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -24,9 +27,35 @@
...
+
+    // Send the JSON as response
+    response.setContentType("application/json;");
+    response.getWriter().println(json);
   }
+
+  /**
+   * Convert to JSON using Gson
+   */
+  private String convertToJson(ArrayList comments) {
```
brettapeters, line 53: Add type parameter. References to generic types should be parameterized.

```java
private String convertToJson(ArrayList<String> comments) {...}
```

aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -14,7 +14,10 @@
 
 package com.google.sps.servlets;
 
+import com.google.gson.Gson;
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
```
brettapeters, line 20: Unused I think

aroquev/src/main/webapp/script.js
```diff
@@ -192,7 +192,21 @@ function getRandomCar() {
...
  */
 function getServletMessage() {
-  fetch('/data').then(response => response.text()).then((greeting)=> {
-    document.getElementById('js-servlet-container').innerHTML = greeting;
+  fetch('/data').then(response => response.json()).then((comments)=> {
+    // Get the cities container
+    const citiesContainer = document.getElementById('js-servlet-container');
+    citiesContainer.innerHTML = '';
+    
+    for (comment of comments) {
```
brettapeters, line 200: `for (let comment of comments) {...}`

Otherwise it's setting `window.comment` each time through the loop.

aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -24,9 +27,35 @@
 @WebServlet("/data")
 public class DataServlet extends HttpServlet {
 
+  private ArrayList<String> comments;
+
+  @Override
+  public void init() {
```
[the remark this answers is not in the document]
seanjohnite: +1. I think it'll make more sense when something is adding comments, but this would probably be the best way to do it.

aroquev/src/main/webapp/script.js
aroquev00: should I use const or let? Const is for when the comment remains unchanged right?

brettapeters: Yes use `const` here. When to use `let` or `const` when either would work is a matter of preference. I think the general consensus is "use const wherever possible", and "choose one way and stay consistent".

aroquev00: Done.

aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
aroquev00: Done.

aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
aroquev00: Done.

aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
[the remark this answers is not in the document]
aroquev00: Mmm, like this?
``` java
private static final ArrayList<String> comments = new ArrayList<String>() {{
    comments.add("I would like to live in NYC.");
    comments.add("Berlin looks like a nice place to have fun.");
    comments.add("I gotta go visit Japan one day!");
  }};  
```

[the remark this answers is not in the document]
aroquev00: I found it a bit weird that it did not allow me to do this:
``` java
private static final ArrayList<String> comments = new ArrayList<String>();
comments.add("I would like to live in NYC.");
comments.add("Berlin looks like a nice place to have fun.");
comments.add("I gotta go visit Japan one day!");
```
Any idea why? I think because I'm calling a method outside a method (the constructor in the commet that does work)?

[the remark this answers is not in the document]
seanjohnite: Right. You can't really do anything outside of methods or functions in the class in Java. I think this kind of has the different ways you could do this in an initializer: https://howtodoinjava.com/java/collections/arraylist/initialize-arraylist/

PR #36, opened 2020-07-01
Comments get request and display

pamelalozano16:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature

## What is the new behavior?

The js does a get request and for every comment it creates an element with the header (subject), comment, and footer (date).

## Other information
Screenshot

### DIFF
--- a/pamelalozano/src/main/webapp/index.html
+++ b/pamelalozano/src/main/webapp/index.html
@@ -198,6 +198,7 @@
 </div>
 </div>
 <div class="myContainer" id="comments">
+    
     <form action="/data" method="POST">
       <p>Comment:</p>
       <input type="text" name="subject" placeholder="Subject">*<br /><br />
@@ -205,9 +206,7 @@
       <br/><br/>
     <input type="submit" />
 
- <div id="comments-section">
- 
-</div>
+    <div id="comments-section"></div>
 
 </div>    
 
--- a/pamelalozano/src/main/webapp/index.html
+++ b/pamelalozano/src/main/webapp/index.html
@@ -194,10 +194,18 @@
 </div>
 </div>
 </div>
-
-
 </div>
-    
+<div class="myContainer" id="comments">
+    <form action="/data" method="POST">
+
+      <p>Comment:</p>
+      <input type="text" name="subject" placeholder="Subject">*<br /><br />
+      <textarea type="text" name="msg" placeholder="Comment"></textarea>*
+      <br/><br/>
+
+    <input type="submit" />
+</div>    
+
 <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.12.9/umd/popper.min.js" integrity="sha384-ApNbgh9B+Y1QKtv3Rn7W3mgPxhU9K/ScQsAP7hUibX39j7fakFPskvXusvfa0b4Q" crossorigin="anonymous"></script>
 <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0/js/bootstrap.min.js" integrity="sha384-JZR6Spejh4U02d8jOt6vLEHfe/JQGiRRSQQxSfFWpi1MquVdAyjUar5+76PVCmYl" crossorigin="anonymous"></script>
 <script src="script.js"></script>
--- a/pamelalozano/src/main/webapp/index.html
+++ b/pamelalozano/src/main/webapp/index.html
@@ -2,13 +2,15 @@
 <html>
   <head>
     <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>Pamela Portfolio</title>
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0/css/bootstrap.min.css" integrity="sha384-Gn5384xqQ1aoWXA+058RXPxPg6fy4IWvTNh0E263XmFcJlSAwiGgFAW/dAiS6JXm" crossorigin="anonymous">
     <script src="https://use.fontawesome.com/eac174ca5f.js"></script>
+    <link rel="stylesheet" href="https://www.w3schools.com/w3css/4/w3.css">
     <link rel="stylesheet" href="style.css">
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
   </head>
-  <body>
+  <body onload="loadComments()">
 <!-- Navbar -->
 <nav class="navbar navbar-expand-lg navbar-light bg-light">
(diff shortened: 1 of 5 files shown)


pamelalozano16:
Wait I don't know why but all the commits until "bug fix with validate comment function" are from my last pull request that is already approved, I did my changes in a different branch so I don't know what happened. There is no need to check those :) If you know what I did wrong please let me know. Thanks!
@brettapeters @seanjohnite

seanjohnite:
Yeah, it has to do with how things are squashed in the upstream repo. I think that you would have needed to create a new fork off of your change PR from the upstream repo to get rid of them? Don't worry about it though, we only really look at the changes for all the commits together (which will just be your most recent changes)

pamelalozano/src/main/webapp/index.html
```diff
@@ -197,13 +199,16 @@ <h3>MongoDB</h3>
...
-
       <p>Comment:</p>
       <input type="text" name="subject" placeholder="Subject">*<br /><br />
       <textarea type="text" name="msg" placeholder="Comment"></textarea>*
       <br/><br/>
-
     <input type="submit" />
+
+ <div id="comments-section">
+
```
seanjohnite, line 209: nit, can put the closing tag on the same line

pamelalozano/src/main/webapp/script.js
```diff
@@ -12,24 +12,71 @@
...
     event.target.closest('.project').classList.toggle("open");
     location.href = "#c-"+event.target.closest('.project').id;
     }
 });
+
+/** Loads existing Comments */
+async function loadComments() {
+  let response = await fetch('/data');
+  let comments = await response.json();
+  if(comments.length==0) {
```
seanjohnite, line 27: nit: formatting spaces

pamelalozano/src/main/webapp/script.js
```diff
@@ -12,24 +12,71 @@
...
+      document.getElementById("comments-section").innerHTML="No comments";
+  } 
+  else {
+    let commentSection = document.getElementById("comments-section");
+    
+    //The newest comments at the top
+    for(let i=comments.length-1; 0<=i; i--){
```
seanjohnite, line 34: You can also do this by reversing the comments in the list first and then using let comment of comments. This would actually change the list, though. Might be okay to do that though!

pamelalozano/src/main/webapp/script.js
```diff
@@ -12,24 +12,71 @@
...
+function createCard(comment) {
+  let newCard = document.createElement('div');
+  newCard.classList.add('w3-card-4');
+  newCard.append(createCardHeader(comment.subject));
+  newCard.append(createCardComment(comment.msg));
+  newCard.append(createCardFooter(comment.date));
+  return newCard;
+}
+
+/** Creates the comment's header. */
```
seanjohnite, line 52: These three functions are all very similar. Can you create a different helper function that takes a few strings and does the same thing? Might be more [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)

brettapeters: If you're planning on header, comment, and footer diverging later, a helper wouldn't be much help.

pamelalozano16: What do you mean by diverging?

pamelalozano/src/main/webapp/script.js
pamelalozano16: Oh okay let me see how can I sort the json in the server side, should I insert the new comments at the beginning at the json instead of at the end (like an unshift but in java)? or use the parameterizing method?

pamelalozano/src/main/webapp/index.html
pamelalozano16: Done ! Thanks

pamelalozano/src/main/webapp/script.js
pamelalozano16: Sorry I always forget, thank you!

pamelalozano/src/main/webapp/script.js
brettapeters: Right now since the HTML for header, comment and footer looks very similar, it might make sense to refactor the functions into one and pass in parameters like

```js
function createThing(outerType, innerType, text) {
  let container = document.createElement(outerType);
  container.classList.add('w3-container');
  let child = document.createElement(innerType);
  child.innerHTML = text;
  container.append(child);
  return container;
}
```
Use it like
```js
const header = createThing('header', 'h3', headerText);
const comment = createThing('div', 'p', commentText);
```

If they are going to be very different later (diverge from each other), a helper would need more if/else and parameters to handle everything. It might get more complicated than just living with a little repetition. Imagine now you want to add image support to comments. The helper would have to change like this (warning - this is a very contrived example).
```js
function createThing(outerType, innerType, text, imgSrc, thingType) {
  let container = document.createElement(outerType);
  container.classList.add('w3-container');
  container.classList.add(containerClass);
  let child = document.createElement(innerType);
  child.innerHTML = text;
  if (thingType === 'comment' && imgSrc !== null) {
    const img = document.createElement('img');
    img.src = imgSrc;
    child.append(img);
  }
  container.append(child);
  return container;
}
```

I hope it's obvious that the helper function could grow into a monster over time.

pamelalozano16: Ooh Okay I understand, well I'm not sure that i'll keep using that format for the comments I wanted to modify the style but I still don't know how but I think I'll definitely use images an more stuff, I think I'll leave it as it is for now, in case I decide to not change the style I'll make a helper function. :) Thanks

pamelalozano/src/main/webapp/script.js
brettapeters: Yes you could insert a new comment at the front of the list. Or you could reverse the list inside `doGet()` and respond with that.

Here's a more complicated way

1. Store comments as a `ArrayList<Comment>` instead of `ArrayList<String>` i.e. don't store them in their JSON format.
2. When `doGet()` is called, you'll need to sort the comments array using `Arrays.sort(...)` and compare the dates using a custom `Comparator<Comment>`, or implement the `Comparable` interface on `Comment` to compare the date field.

Imagine you also have a field called `upvotes` on `Comment`. You could change the sort comparator to compare `upvotes` instead and show the highest upvoted comments first.

pamelalozano16: Ohh okay okay I think I'll do this as a feature later on, when I already have the datastore. Thank you!

brettapeters: Sounds good. The datastore client lets you specify things like sort-by and limit on queries, so you wouldn't have to manually sort like this.

PR #41, opened 2020-07-02
Load comments from Datastore

aroquev00:
The webapp now stores and loads all comments from Datastore, providing persistence to the data. The doGet method of the servlet was modified to query the comments from DS and send them to the client.

Link to deployment: https://aroquev-step-2020.uc.r.appspot.com/

### DIFF
--- a/aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
+++ b/aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
@@ -14,6 +14,9 @@
 
 package com.google.sps.servlets;
 
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
 import com.google.gson.Gson;
 import java.io.IOException;
 import java.util.ArrayList;
@@ -42,6 +45,15 @@ public class DataServlet extends HttpServlet {
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
     // probaby verify that message is not empty
     String comment = request.getParameter("comments-body-input");
+    long timestamp = System.currentTimeMillis();
+
+    Entity commentEntity = new Entity("Comment");
+    commentEntity.setProperty("commentBody", comment);
+    commentEntity.setProperty("timestamp", timestamp);
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    datastore.put(commentEntity);
+
     comments.add(comment);
     response.sendRedirect("/");
   }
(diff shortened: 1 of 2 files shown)


aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -27,12 +30,25 @@
...
+    Query query = new Query("Comment").addSort("timestamp", SortDirection.DESCENDING);
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery results = datastore.prepare(query);
+
+    final ArrayList<String> comments = new ArrayList<String>();
+    for (Entity entity : results.asIterable()) {
+      long id = entity.getKey().getId();
+      String comment = (String) entity.getProperty("commentBody");
+      // Timestamp is not used at the moment, but will help in next feature
```
brettapeters, line 46: `id` is also not used. You should remove the unused variables until they're needed.

aroquev00: Done.

aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -27,12 +30,22 @@
 
 /** Servlet that returns some example content. TODO: modify this file to handle comments data */
```
seanjohnite, line 31: update TODO?

aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -27,12 +30,22 @@
...
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    Query query = new Query("Comment").addSort("timestamp", SortDirection.DESCENDING);
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery results = datastore.prepare(query);
+
+    final ArrayList<String> comments = new ArrayList<String>();
+    for (Entity entity : results.asIterable()) {
+      String comment = (String) entity.getProperty("commentBody");
```
seanjohnite, line 44: Do these comments have any metadata? Are they just strings?

aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
aroquev00: Done.

aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
aroquev00: As for now they are, but will add info such as timestamp and other things and for that I plan to create a Comment class.

PR #45, opened 2020-07-02
Add number of comments to display

aroquev00:
The user can now insert a number for the maximum number of comments to be displayed. 
If valid, The number is sent as a query string to the servlet and the servlet queries that number of comments and sends them to the client. If not valid, the default number of comments is 5.

image

image

### DIFF
--- a/aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
+++ b/aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
@@ -36,7 +36,6 @@ public class DataServlet extends HttpServlet {
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     int maxComments = Integer.parseInt(request.getParameter("maxComments"));
-    System.out.println(maxComments);
 
     Query query = new Query("Comment").addSort("timestamp", SortDirection.DESCENDING);
 
--- a/aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
+++ b/aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
@@ -14,6 +14,9 @@
 
 package com.google.sps.servlets;
 
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
 import com.google.gson.Gson;
 import java.io.IOException;
 import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


aroquev/src/main/webapp/index.html
```diff
@@ -185,6 +185,14 @@ <h2>Comments!</h2>
       </form>
   
       <h3>Previous comments</h3>
+
+      <label for="maxComments">Change number of comments displayed (between 1 and 10):</label>
+      <input type="number" id="maxComments" name="quantity" min="1" max="10" value="5" onchange="getServletComments()">
```
aroquev00, line 190: Do you think I should add this inside a form so that I can check in HTML the min and max values? Or is it okay if I do it on the JS side? I read that HTML is faster.

aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -34,13 +35,16 @@
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    int maxComments = Integer.parseInt(request.getParameter("maxComments"));
+    System.out.println(maxComments);
```
seanjohnite, line 39: remove?

aroquev/src/main/webapp/index.html
seanjohnite: What do you mean by add inside a form?

aroquev00: Mmm so I apparently if I place the <input> tag inside a <form> tag, HTML can ensure that the number inserted is within the range, so I would not have to use JS to check this.
Like this
image
But I find it odd that I would be using a <form> without sending a POST request anywhere...

pamelalozano16: I think @brettapeters mentioned all the filtration, limits, etc. should be done in the server side, so the js and html just display the data. I did some work with cursors I still have some questions but maybe I can help you :)
Here's the link in case you want to check them out: https://cloud.google.com/appengine/docs/standard/java/datastore/query-cursors

aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
aroquev00: Done.

aroquev/src/main/webapp/index.html
aroquev00: Ohhh okay, thanks @pamelalozano16 !!!

brettapeters: You can trigger html input validation manually and even set custom validation messages if you wish. I made a small demo here https://codepen.io/bap13/pen/NWxYqYZ.

You can also use the CSS psuedo-class `:invalid` to give your input extra styling when it's invalid.

aroquev00: Cool! I will implement it.

PR #46, opened 2020-07-02
Delete all comments

aroquev00:
Add button that deletes all comments on the webpage.
The button triggers a JS function that sends a POST request to the DeleteData.java servlet, which queries all comments and deletes them one by one. The JS function waits until it gets a response and then refreshes the comments section.

### DIFF
--- a/aroquev/src/main/java/com/google/sps/servlets/DeleteServlet.java
+++ b/aroquev/src/main/java/com/google/sps/servlets/DeleteServlet.java
@@ -1,0 +1,50 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.servlets;
+
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.Query.SortDirection;
+import com.google.appengine.api.datastore.Key;
+import com.google.gson.Gson;
+import java.io.IOException;
+import java.util.ArrayList;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that returns some example content.*/
+@WebServlet("/delete-data")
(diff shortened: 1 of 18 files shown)


aroquev/src/main/webapp/script.js
```diff
@@ -212,9 +214,39 @@ function getServletMessage() {
...
   return liElement;
 }
+
+/**
+ * Deletes all comments in DS and then refreshes comments section
+ */
+function deleteAllComments() {
+  fetch('/delete-data', {method: 'POST'}).then(() => {
+    getServletComments();
+  });
```
aroquev00, line 251: Here the response from the server gets sent once it has executed all of its code? Or is it possible that it returns a response while it has not yet deleted all comments?

seanjohnite: Yes, I think that the server should only respond when the request handler has finished executing, so all comments should be deleted in this case. Do you see something else?

aroquev00: Nope, just in some random cases the page reloads and still displays some comments, but this rarely happens and only in my dev-server.

aroquev/src/main/java/com/google/sps/servlets/DeleteServlet.java
```diff
@@ -0,0 +1,50 @@
...
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    Query query = new Query("Comment");
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery results = datastore.prepare(query);
+    
+    Key commentKey;
+    for (Entity entity : results.asIterable()) {
```
brettapeters, line 44: I guess there isn't a way to do a batch delete with the App Engine API for Datastore, but normally you'd want to do a batch request instead of several blocking delete requests in a loop.

aroquev/src/main/webapp/script.js
brettapeters: It could have something to do with the way Datastore works. It's not guaranteed that a read immediately following a write will return consistent data. You can read more about this here https://cloud.google.com/datastore/docs/concepts/structuring_for_strong_consistency

brettapeters: That link applies to the older datastore. Your project should be using "Firestore in Datastore mode", which has strong consistency.

aroquev00: Mmm yep you're right, so all our queries should have strong consistency?

aroquev/src/main/java/com/google/sps/servlets/DeleteServlet.java
aroquev00: This should be it right? https://cloud.google.com/appengine/docs/standard/java/datastore/creating-entities#Batch_operations
Because these are all the same Entities this shouldn't be an issue?
A batch put() or delete() call may succeed for some entities but not others. If it is important that the call succeed completely or fail completely, use a transaction with all affected entities in the same entity group.

aroquev/src/main/webapp/script.js
brettapeters: That should be the case. If you read before a write is finished you'd get stale data - with or without strong consistency.

aroquev/src/main/java/com/google/sps/servlets/DeleteServlet.java
brettapeters: Oh yeah... I didn't read that right. You should use the batch-style call by getting all the keys then call `datastore.delete(keys);` once. And use a transaction if you want to abort if some deletes fail.

PR #38, opened 2020-07-01
PR for walkthrough #4 of Week 3

aroquev00:
Comments can now be posted to the webpage. A comment is sent to the servlet with POST operation, and page is reloaded. JS now fetches the comments and displays them as a list in the comments section.

image
No comments yet

image
Example of posted comments

PR #47, opened 2020-07-03
Pagination in comments

pamelalozano16:
#38  PR Type

- [ ] Bugfix
- [x] Feature

## What is the new behavior?

The servlet now makes the get request according to the parameters, if there's an existing cursor that means that the user is requesting a next page of the comments or to return to the first one. And the get "/data" now returns an object with the comments and the cursor of the last comment from the data, so if the user clicks next the elements in the DOM will change to the next page of the request (that know used the cursor).

I think I can merge the nextPage() and backPage() with a helper function but since I'm calling it from the onclick on the html I'm not sure how to pass arguments.

I deployed the changes to the app since it's from this week's walkthrough so you can watch it there.

I tried to document enough but if you have any questions let me know :)

### DIFF
--- a/pamelalozano/src/main/java/com/google/sps/servlets/DataServlet.java
+++ b/pamelalozano/src/main/java/com/google/sps/servlets/DataServlet.java
@@ -46,17 +46,14 @@ public class DataServlet extends HttpServlet {
     FetchOptions fetchOptions = FetchOptions.Builder.withLimit(3);
 
      String startCursor = request.getParameter("cursor");
-     String next = request.getParameter("next"); //if back this is null
-     String back = request.getParameter("back"); //if next this is null
+     String pageDirection = request.getParameter("pageDirection");
 
     //If the request is for a change of page
-    if (startCursor != null && next != null) {
-        
+    if (startCursor != null && "next".equals(pageDirection)) {
         //Next 3 comments
       fetchOptions.startCursor(Cursor.fromWebSafeString(startCursor));
     
-    } else if (startCursor != null && back != null) {
-        
+    } else if (startCursor != null && "back".equals(pageDirection)) {
         //Back to the first 3 comments 
         //I couldn't figure out how to get the getStartCursor()
       fetchOptions.endCursor(Cursor.fromWebSafeString(startCursor));
--- a/pamelalozano/src/main/java/com/google/sps/servlets/DataServlet.java
+++ b/pamelalozano/src/main/java/com/google/sps/servlets/DataServlet.java
@@ -14,6 +14,9 @@
 
 package com.google.sps.servlets;
 
+import com.google.appengine.api.datastore.Cursor;
+import com.google.appengine.api.datastore.QueryResultList;
+import com.google.appengine.api.datastore.FetchOptions;
 import com.google.appengine.api.datastore.PreparedQuery;
 import com.google.appengine.api.datastore.Query;
 import com.google.appengine.api.datastore.Query.SortDirection;
@@ -21,6 +24,7 @@ import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
 import com.google.appengine.api.datastore.Entity;
 import java.util.ArrayList; 
+import java.util.List; 
 import java.util.Date; 
 import java.io.IOException;
 import javax.servlet.annotation.WebServlet;
@@ -42,10 +46,11 @@ public class DataServlet extends HttpServlet {
     ArrayList<String> comments = new ArrayList<>();  
 
     Query query = new Query("Comment").addSort("date", SortDirection.DESCENDING);
-    PreparedQuery results = datastore.prepare(query);
+    PreparedQuery resultsQuery = datastore.prepare(query);
+    List<Entity> results = resultsQuery.asList(FetchOptions.Builder.withLimit(3));
 
     //Query of entities to JSON
-    for (Entity entity : results.asIterable()) {
+    for (Entity entity : results) {
         
         //Returns comment converted to json (method in Comment object)
         String comment = entityToString(entity);
--- a/pamelalozano/src/main/java/com/google/sps/servlets/DataServlet.java
+++ b/pamelalozano/src/main/java/com/google/sps/servlets/DataServlet.java
@@ -14,6 +14,9 @@
 
 package com.google.sps.servlets;
 
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.Query.SortDirection;
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
 import com.google.appengine.api.datastore.Entity;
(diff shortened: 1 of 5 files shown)


pamelalozano/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -14,31 +14,99 @@
...
+    //If the request is for a change of page
+    if (startCursor != null && next != null) {
+        
+        //Next 3 comments
+      fetchOptions.startCursor(Cursor.fromWebSafeString(startCursor));
+    
+    } else if (startCursor != null && back != null) {
+        
+        //Back to the first 3 comments 
+        //I couldn't figure out how to get the getStartCursor()
```
seanjohnite, line 62: You may be able to use QueryResultIterator instead QueryResultList to get both the start and end cursors? https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/QueryResultIterator.html

pamelalozano/src/main/webapp/script.js
```diff
@@ -18,23 +18,90 @@ document.getElementById("projects").addEventListener("click", function (event) {
...
+    comments.forEach(comment => {
+        console.log(comment);
+        let newCard = createCard(comment);
+        commentSection.append(newCard);
+    });
+
+  };
+}
+
+async function nextPage(){
```
seanjohnite, line 46: Since these are so similar, you can create a helper function that each of these functions can call to avoid repeating code.

pamelalozano/src/main/java/com/google/sps/servlets/DataServlet.java
pamelalozano16: Yeah I spent a long time trying to implement this but i don't know why I get the error "Class query doesn't take parameters" I guess because the query is declared like this (according to https://cloud.google.com/datastore/docs/concepts/queries):
Query<Entity> query = Query.newEntityQueryBuilder()

But it didn't let me so I tried to look for another import I may need but I didn't find anything or what I found didn't work. 
The class Query and StructuredQuery seem contain the getStartCursor method:
https://developers.google.com/resources/api-libraries/documentation/datastore/v1/java/latest/com/google/api/services/datastore/v1/model/Query.html
https://googleapis.dev/java/google-cloud-clients/0.90.0-alpha/com/google/cloud/datastore/StructuredQuery.html

Let me know if you find out why this is happening :)

pamelalozano/src/main/webapp/script.js
pamelalozano16: Yeah I actually mentioned this in the PR description, it's done ! :)

pamelalozano/src/main/java/com/google/sps/servlets/DataServlet.java
brettapeters: Not sure but maybe you want to try setting the start cursor the same, but call `Query.reverse()`?https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/Query.html#reverse--. Results might be sorted in the opposite way though.

I think you're looking at the docs for the newer API, which they're not having you use for some reason. You want to look through https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/package-summary

pamelalozano/src/main/webapp/script.js
```diff
@@ -18,25 +18,71 @@ document.getElementById("projects").addEventListener("click", function (event) {
...
-      document.getElementById("comments-section").innerHTML="No comments";
+      document.getElementById("comments-list").innerHTML="No comments";
   } 
   else {
-    let commentSection = document.getElementById("comments-section");
+    let commentSection = document.getElementById("comments-list");
     
     //The newest comments at the top
     comments.forEach(comment => {
+        console.log(comment);
```
brettapeters, line 38: nit. remove `console.log`

pamelalozano/src/main/webapp/script.js
```diff
@@ -18,25 +18,71 @@ document.getElementById("projects").addEventListener("click", function (event) {
...
         let newCard = createCard(comment);
         commentSection.append(newCard);
     });
 
   };
 }
 
+/** Changes comment page */
+async function changePage(type){
+    let reponse;
```
brettapeters, line 48:
```suggestion
    let response;
```

pamelalozano/src/main/webapp/index.html
```diff
@@ -224,9 +224,12 @@ <h3>MongoDB</h3>
 
     <div id="comments-section">
       <p style="text-align: center;">
-        <i class="fa fa-chevron-circle-left fa-2x" aria-hidden="true" style="float: left;"></i>
+
+        <i class="fa fa-chevron-circle-left fa-2x" aria-hidden="true" style="float: left; cursor: pointer;" onclick="changePage('back')"></i>
```
brettapeters, line 228: You could hide the "back" icon if the current page is 1.

Also hide the "next" icon if there are no more results. I think that would be a little harder to figure out.

pamelalozano/src/main/webapp/script.js
```diff
@@ -18,25 +18,71 @@ document.getElementById("projects").addEventListener("click", function (event) {
...
+            }
+        } 
+        else if (type == 'next') {
+            pageNum++;
+        }
+        document.getElementById("pageNumber").innerText= pageNum;
+
+        //The newest comments at the top
+        comments.forEach(comment => {
+        console.log(comment);
```
brettapeters, line 79: nit. remove `console.log`

pamelalozano/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -43,23 +43,63 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
 
-    ArrayList<String> comments = new ArrayList<>();  
+    FetchOptions fetchOptions = FetchOptions.Builder.withLimit(3);
 
-    Query query = new Query("Comment").addSort("date", SortDirection.DESCENDING);
-    PreparedQuery resultsQuery = datastore.prepare(query);
-    List<Entity> results = resultsQuery.asList(FetchOptions.Builder.withLimit(3));
+     String startCursor = request.getParameter("cursor");
+     String next = request.getParameter("next"); //if back this is null
```
brettapeters, line 49: This could be a single parameter "paginationDirection" which could have a value of "up" or "down". The URL param would then be `?paginationDirection=up` for newer comments; `?paginationDirection=down` for older comments.

seanjohnite:
I think this LG Pamela, let us know when you want to merge it.

pamelalozano/src/main/java/com/google/sps/servlets/DataServlet.java
pamelalozano16: Ohh okay perfect I think I'll try that when I'm done with week 4 so I don't fall behind. Thank you!

pamelalozano/src/main/webapp/index.html
pamelalozano16: Sure I guess I can toggle the class according to the data response but I think I'll do it when I'm done with week 4 :) Thank you!

PR #11, opened 2020-06-25
Add Personal Info and Projects

aroquev00:
Added a personal introduction, photo, personal background and projects list.

No styling added yet, will add as soon as all info is on HTML file.

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] PR has a descriptive title
- [x] PR contains closely related changes only (should not have many unrelated changes)
- [ ] STEP pod members have been added as reviewers
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## What is the new behavior?

image

image

PR #48, opened 2020-07-04
Comment username, date and votes

aroquev00:
The comments now include details such as who posted it, the day and time when it was posted, and the popularity of it. The popularity is defined as the difference between the upvotes and downvotes, and allows the user to vote up or down by clicking the appropriate button. 

## PR Checklist

- [x] PR has a descriptive title
- [x] PR contains closely related changes only (should not have many unrelated changes)
- [ ] STEP pod members have been added as reviewers

## PR Type

- [ ] Bugfix
- [x] Feature

Nonexistent feature 

Issue Number: N/A

## What is the new behavior?
The comments now include details such as who posted it, the day and time when it was posted, and the popularity of it. The popularity is defined as the difference between the upvotes and downvotes, and allows the user to vote up or down by clicking the appropriate button. 

image

## Other information
Closes https://github.com/aroquev00/step191-2020/issues/13
Closes https://github.com/aroquev00/step191-2020/issues/10
Closes https://github.com/aroquev00/step191-2020/issues/11
Closes https://github.com/aroquev00/step191-2020/issues/16


aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -55,13 +67,16 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
 
     Entity commentEntity = new Entity("Comment");
+    commentEntity.setProperty("commentUsername", username);
     commentEntity.setProperty("commentBody", comment);
     commentEntity.setProperty("timestamp", timestamp);
+    commentEntity.setProperty("upvotes", 0);
+    commentEntity.setProperty("downvotes", 0);
 
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
     datastore.put(commentEntity);
```
seanjohnite, line 82: May not need redirect now?

aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -34,14 +36,24 @@
...
 
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
     PreparedQuery results = datastore.prepare(query);
 
-    final ArrayList<String> comments = new ArrayList<String>();
-    for (Entity entity : results.asIterable()) {
-      String comment = (String) entity.getProperty("commentBody");
+    final ArrayList<Comment> comments = new ArrayList<Comment>();
+    for (Entity entity : results.asIterable(FetchOptions.Builder.withLimit(maxComments))) {
+      long id = entity.getKey().getId();
```
seanjohnite, line 48: It might be nice to create a Comment factory that can take an Entity comment and create a Comment? Could be used here and a few other places.

aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -55,13 +67,16 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
 
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    // TODO: verify that message is not empty
+    String username = request.getParameter("comments-username-input");
     String comment = request.getParameter("comments-body-input");
     long timestamp = System.currentTimeMillis();
 
     Entity commentEntity = new Entity("Comment");
```
seanjohnite, line 74: Comment class could also have a "toEntity" method to allow putting into Datastore.

aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
brettapeters: +1 🏭 
And/or use the Builder pattern so you don't have a constructor with a bunch of parameters.

aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
brettapeters: +1

aroquev/src/main/java/com/google/sps/servlets/VotesServlet.java
```diff
@@ -0,0 +1,80 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that updates upvotes or downvotes of a comment*/
+@WebServlet("/vote-comment")
+public class VotesServlet extends HttpServlet { 
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String jsonString = deserializeJson(request);
```
brettapeters, line 37: There is an overload of `toJson` that takes a `java.io.Reader` directly so you don't need the intermediate conversion of the request body to a string. `HttpServletRequest#getReader()` returns a `java.io.BufferedReader`, which extends `java.io.Reader`.

`JsonObject jsonObj = new Gson().fromJson(request.getReader(), JsonObject.class);`

Shazam. one-liner.

aroquev/src/main/webapp/script.js
```diff
@@ -192,29 +192,134 @@ function getRandomCar() {
...
+  const popularity = document.createElement('div');
+  popularity.innerText = `Popularity: ${comment.upvotes - comment.downvotes}`;
+
+  const body = document.createElement('div');
+  body.innerText = comment.body;
+  body.classList.add('comment-body');
+
+  const upvote = document.createElement('button');
+  upvote.innerText = "Upvote";
+  upvote.addEventListener('click', () => {
```
brettapeters, line 269: Like I mentioned in the sync, the event listener could be removed after one click as a very simple way to deter spamming votes. Wouldn't stop users from just refreshing or writing a loop in the console, but it's better than nothing.

aroquev/src/main/java/com/google/sps/data/Comment.java
```diff
@@ -0,0 +1,34 @@
...
+
+package com.google.sps.data;
+
+public final class Comment {
+
+  private final long id;
+  private final String body;
+  private final String username;
+  private final long timestamp;
+  private final int upvotes;
```
brettapeters, line 23: I think if you wanted to enforce a user can only upvote/downvote once per comment, you'd want to store a "Vote" as a separate entity. That's probably too much work for the starter project though.

seanjohnite:
Let us know when you want to merge this in Armando!

/giphy BOOM

aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
aroquev00: This redirect? 
`response.sendRedirect("/");`

aroquev/src/main/java/com/google/sps/servlets/VotesServlet.java
aroquev00: Awesome, done.

aroquev/src/main/webapp/script.js
aroquev00: Mmm I see. The thing is that because when the user hits the upvote comment, the getComments() method gets called and resets the whole section, so if I disable the clicking it gets enabled instantly because it's like the user never clicked anything. I did this to keep the DS as the source of data at all times, but now that's not helping in this situation. I was planning on kinda fixing this with the login auth, that would allow the user to either upvote or downvote the comment once.

aroquev/src/main/java/com/google/sps/data/Comment.java
aroquev00: Okay, if I will try to implement it if I have the chance. As for now, do you think there is a good approach I could take given the fact that once anything is clicked on everything gets loaded once again?

aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
aroquev00: Okay so this method would mean that I have a Comment object which can call this toEntity() method and return an Entity that represents that Comment right?

aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
brettapeters: Correct. Most times I've seen POST request handlers (that create a new thing) return the thing that was created.

brettapeters:
Please fix build failure so we can merge.

aroquev00:
https://tenor.com/wHOT.gif
image

aroquev/src/main/java/com/google/sps/data/Comment.java
```diff
@@ -0,0 +1,63 @@
...
+  public Comment(long id, String body, String username, long timestamp, int upvotes, int downvotes) {
+    this.id = id;
+    this.body = body;
+    this.username = username;
+    this.timestamp = timestamp;
+    this.upvotes = upvotes;
+    this.downvotes = downvotes;
+  }
+
+  public Comment(Entity entity) {
```
brettapeters, line 37: This PR is getting ooold so we should merge, but FYI the general consensus is to use static factory methods over constructors for this type of thing. The main advantage in this case is the you would have a name that clearly communicates what it's for.

```java
public static Comment fromEntity(Entity entity) {
  // put properties into local variables
  return new Entity(id, body, ...);
}
```

PR #50, opened 2020-07-06
Implement Brett's Input validation

aroquev00:
Implemented Brett's numeric input validation with CSS and JS.

image

### DIFF
--- a/aroquev/src/main/webapp/script.js
+++ b/aroquev/src/main/webapp/script.js
@@ -223,11 +223,12 @@ function verifyNumberCommentsInput() {
   let maxComments = 5;
 
   let maxCommentsInput = document.getElementById("maxComments").value;
+  const commentsErrorContainer = document.getElementById('js-comments-input-error-container');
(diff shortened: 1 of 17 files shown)


aroquev/src/main/webapp/script.js
```diff
@@ -221,15 +221,11 @@ function getServletComments() {
 function verifyNumberCommentsInput() {
   // Default value
   let maxComments = 5;
-
-  let maxCommentsInput = document.getElementById("maxComments").value;
-  const commentsErrorContainer = document.getElementById('js-comments-input-error-container');
+  let maxCommentsContainer = document.getElementById("comment-number-input");
+  maxCommentsContainer.reportValidity();
```
brettapeters, line 225: Maybe you want to prevent the request to get comments if the number is invalid. As far as I can tell, if a user types in "20", they will see an error and then 5 comments will be fetched anyway. This might be non-obvious behavior to the user.

aroquev00: Gotcha, will implement that

PR #53, opened 2020-07-07
Auth object to display info in "/user"

pamelalozano16:
Add email to auth object

## PR Type

- [ ] Bugfix
- [x] Feature

## What is the new behavior?

Auth object to display info in "/user" and use it in DOM

## Other information
Screenshot 2020-07-07 at 15 36 59 - Display 1
Screenshot 2020-07-07 at 15 36 36 - Display 1

### DIFF
--- a/pamelalozano/src/main/java/com/google/sps/servlets/Auth.java
+++ b/pamelalozano/src/main/java/com/google/sps/servlets/Auth.java
@@ -1,0 +1,63 @@
+
+package com.google.sps.servlets;
+
+
+import java.util.*;  
+import com.google.gson.Gson;
+import java.io.IOException;
+
+public class Auth {
+  
+  private boolean isLoggedIn;
+  private boolean isAdmin;
+  private String logoutUrl;
+  private String loginUrl;
+  private String activeUser;
+
+  public Auth(boolean isLoggedIn) {
+    this.isLoggedIn = isLoggedIn;
+    this.isAdmin = false;
+    this.logoutUrl = null;
+    this.loginUrl = null;
+    this.activeUser = null;
+  }
+  
+  public void setLoginUrl(String loginUrl) {
+      this.loginUrl = loginUrl;
+      this.logoutUrl = null;
+      this.activeUser = null;
+  }
+  public void setLogoutUrl(String logoutUrl) {
+      this.logoutUrl = logoutUrl;
+      this.loginUrl = null;
+  }
+  public void setAdmin(boolean isAdmin) {
+      this.isAdmin = isAdmin;
+  }
+  public void setIsLoggedIn(boolean isLoggedIn) {
+      this.isLoggedIn = isLoggedIn;
+  }
+  public void setActiveUser(String activeUser) {
+      this.activeUser = activeUser;
+  }
+  public String getLoginUrl() {
+      return this.loginUrl;
+  }
+  public String getLogoutUrl() {
+      return this.logoutUrl;
+  }
+  public String getUserEmail() {
+      return this.activeUser;
+  }
+  public Boolean isAuthenticated() {
+      return this.isLoggedIn;
+  }
+  public Boolean isAdministrator() {
+      return this.isAdmin;
+  }
+  public String toJson() {
+    Gson gson = new Gson();
+    String json = gson.toJson(this);
+    return json;
+  }
+};
--- a/pamelalozano/src/main/java/com/google/sps/servlets/Auth.java
+++ b/pamelalozano/src/main/java/com/google/sps/servlets/Auth.java
@@ -1,8 +1,8 @@
-
 package com.google.sps.servlets;
 
-
-import java.util.*;  
+import java.lang.String;  
+import java.lang.Boolean;  
+import java.util.Date;  
 import com.google.gson.Gson;
 import java.io.IOException;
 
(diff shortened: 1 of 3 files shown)


pamelalozano/src/main/java/com/google/sps/servlets/Auth.java
```diff
@@ -0,0 +1,63 @@
+
```
seanjohnite, line 1: nit: space?

pamelalozano/src/main/java/com/google/sps/servlets/Auth.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+  public Boolean isAdministrator() {
+      return this.isAdmin;
+  }
+  public String toJson() {
+    Gson gson = new Gson();
+    String json = gson.toJson(this);
+    return json;
+  }
+};
```
seanjohnite, line 63: nit: no semicolon needed here

pamelalozano/src/main/java/com/google/sps/servlets/Auth.java
```diff
@@ -0,0 +1,63 @@
+
+package com.google.sps.servlets;
+
+
+import java.util.*;  
+import com.google.gson.Gson;
+import java.io.IOException;
+
+public class Auth {
```
seanjohnite, line 9: FYI mostly, as this is totally fine as is:

Best practice here would be to make all of these final fields. The way to do this would be to create a Auth.Builder that can add all of the fields individually and then call ".build()" to actually create the Auth instance. At Google, we would use a code generation annotation to do this called AutoValue. Here are the docs on how to use it! https://github.com/google/auto/blob/master/value/userguide/index.md

pamelalozano/src/main/java/com/google/sps/servlets/Auth.java
```diff
@@ -0,0 +1,63 @@
+
+package com.google.sps.servlets;
+
+
+import java.util.*;
```
seanjohnite, line 5: try to make these specific imports

pamelalozano/src/main/java/com/google/sps/servlets/Auth.java
pamelalozano16: Emm does this means I should delete it? Or add more space? haha

pamelalozano/src/main/java/com/google/sps/servlets/Auth.java
pamelalozano16: Oh true done

pamelalozano/src/main/java/com/google/sps/servlets/Auth.java
pamelalozano16: So you mean something like this right?
https://www.vogella.com/tutorials/DesignPatternBuilder/article.html

brettapeters: Hand written builder example. This is very boring & error-prone to write so we can make AutoValue generate this. 

```java
public class Auth {
  private final boolean isLoggedIn;
  // More private final fields

  // Public getter for isLoggedIn  
  public boolean getIsLoggedIn() {
    return isLoggedIn;
  }
  // Other getters. No setters!

  // Auth constructor is private.
  // There is no way to create an instance of Auth without using the Builder
  private Auth(Builder builder) {
    isLoggedIn = builder.isLoggedIn;
    // Set other Auth final fields from builder 
  }

  // static inner class can be instantiated like `new Auth.Builder()`
  public static class Builder {
    private isLoggedIn;
    // Other private (not final) fields internal to Builder
    
    public Builder setIsLoggedIn(boolean isLoggedIn) {
      this.isLoggedIn = isLoggedIn;
      return this;
    }

    // Other "set..." methods to set internal Builder fields

    public Auth build() {
      return new Auth(this);
    }
  }
}
```

Usage
```java
Auth myAuth = new Auth.Builder().setIsLoggedIn(true).build();
System.out.println("User logged in?: " + myAuth.getIsLoggedIn());
```

Then the obvious issue is "What if I have an instance of `Auth` but I need to update one value?". You write a method on Auth `Builder toBuilder() {...}` that returns a new builder with all of the fields copied and use it like this 
```java
Auth updatedAuth = oldAuth.toBuilder().setWhatever(newValue).build();
```

AutoValue can also generate `toBuilder()` for you.

brettapeters: Becomes this when using AutoValue.

```java
import com.google.auto.value.AutoValue;

@AutoValue
abstract class Auth {
  abstract boolean isLoggedIn();

  // So you can get a builder like `Auth.builder()` instead of `new AutoValue_Auth.Builder()`
  static Builder builder() {
    return new AutoValue_Auth.Builder();
  }

  @AutoValue.Builder
  abstract static class Builder {
    abstract Builder setIsLoggedIn(boolean isLoggedIn);
    abstract Auth build();
  }
}
```

pamelalozano16: Thank you so much Brett this is super helpful!

pamelalozano16: Done!

seanjohnite:
Nice!!

PR #55, opened 2020-07-08
Styling

aroquev00:
This one is mostly about styling, so here's the link to the deployment for you guys to explore it.
Any comments about how to make it look cooler are more than welcome!
https://aroquev-step-2020.uc.r.appspot.com/#


aroquev/src/main/webapp/index.html
```diff
@@ -5,9 +5,21 @@
...
+  <link href="https://fonts.googleapis.com/css2?family=Noto+Sans+JP&display=swap" rel="stylesheet">
+  <link rel="icon" 
+      type="image/png" 
+      href="images/noogler-hat.png">
 </head>
 
 <body onload="getServletComments()">
+  <div class="topnav">
+    <a class="active" href="#">Home</a>
+    <a href="#about-me">About Me</a>
```
brettapeters, line 17: You could make these into separate pages (about.html, projects.html, etc.). You aren't constrained to having every section inside index.html.

aroquev/src/main/webapp/index.html
```diff
@@ -195,20 +207,22 @@ <h3>Previous comments</h3>
...
-      <p><a id="linkedin-link"
+      <a id="github-link" href="https://github.com/aroquev00" target="_blank"><img src="images/github-logo.png"></a>
+      <a id="linkedin-link"
           href="https://mx.linkedin.com/in/armando-roque-a533a316b/en-us?trk=people-guest_people_search-card"
-          target="_blank">Click here to see my LinkedIn Profile! <img src="images/Linkedin-logo.png"></a></p>
+          target="_blank"><img src="images/Linkedin-logo.png"></a>
     </div>
-
-
+    <br>
```
brettapeters, line 220: Use margin instead of `<br>` if your goal is to increase space between elements.

`<br>` is mainly used to force line breaks in text.

aroquev/src/main/webapp/style.css
```diff
@@ -1,3 +1,36 @@
...
+/* For the header */
+/* Add a black background color to the top navigation */
+.topnav {
+  background-color: #333;
+  overflow: hidden;
+}
+
+/* Style the links inside the navigation bar */
+.topnav a {
+  float: left;
```
brettapeters, line 14: I don't know if you really need to use `float` for this layout. You can get the same thing if you give the parent div `display: flex` and `flex-wrap: wrap` if you still want a nav link to flow down to the next line on small viewports.

aroquev/src/main/webapp/style.css
```diff
@@ -1,3 +1,36 @@
...
+}
+
+/* Change the color of links on hover */
+.topnav a:hover {
+  background-color: #ddd;
+  color: black;
+}
+
+/* Add a color to the active/current link */
+.topnav a.active {
```
brettapeters, line 29: You'll need to have some JS that applies this class to the nav item that is clicked. Right now the "active" item is always "Home"

aroquev/src/main/webapp/style.css
```diff
@@ -124,6 +161,58 @@
   to {opacity: 1}
 }
 
+/* Footer styling */
+footer {
+  text-align: center;
+  padding: 3px;
+  background-color: #4397ac;
+  color: white;
+  display: block;
```
brettapeters, line 170: `<footer>` is block-level by default

aroquev/src/main/webapp/index.html
aroquev00: True, it's starting to look a bit too cramped.

aroquev/src/main/webapp/index.html
aroquev00: Done, removed `<br>` tags and did the effect by setting margins in CSS rules.

aroquev/src/main/webapp/style.css
aroquev00: Right, I left it like that because I thought red looked cool on the Home button hehe, but once I divide it in more pages, it will change to the page you're in.

aroquev/src/main/webapp/style.css
aroquev00: Done.

aroquev/src/main/webapp/style.css
aroquev00: So I just added those properties to the .topnav rule?

PR #56, opened 2020-07-08
Comments logged in

aroquev00:
Now you can only post comments if you are logged in. But even if you are not, you can still see the comments section (probably should also limit the voting capabilities). The way it works is JS fetches the login servlet and returns the login status of user and the corresponding login/logout link, but I think Pamela's approach of creating a Auth class is better.

Link to deployment for demo (remember to use Incognito to avoid logging out of G account): https://aroquev-step-2020.uc.r.appspot.com/


aroquev/src/main/webapp/script.js
```diff
@@ -346,3 +346,86 @@ function deleteAllComments() {
...
+    const pElement = document.createElement('p');
+    pElement.innerText = "You are welcome to leave a comment if you found something interesting or just have anything to say!";
+    
+    const bodyTextArea = document.createElement('textarea');
+    bodyTextArea.id = 'comments-body-input';
+    bodyTextArea.name = 'comments-body-input';
+    bodyTextArea.required = 'required';
+    bodyTextArea.placeholder = 'Write your comment here';
+
+    const usernameTextArea = document.createElement('textarea');
```
seanjohnite, line 381: Seems this username input isn't used anymore?

aroquev/src/main/webapp/script.js
```diff
@@ -346,3 +346,86 @@ function deleteAllComments() {
...
+  const writeCommentsBox = document.getElementById('js-write-comment-box');
+  const loginInfo = await checkLoggedIn();
+  if (loginInfo.isLoggedIn) {  
+    const commentForm = document.createElement('form');
+    commentForm.action = 'javascript:;';
+    commentForm.addEventListener('submit', () => {
+      submitComment(commentForm);
+    });
+
+    // Build elements inside form
```
seanjohnite, line 371: This might be a slick solution to creating a lot of static HTML content!

https://stackoverflow.com/questions/494143/creating-a-new-dom-element-from-an-html-string-using-built-in-dom-methods-or-pro/35385518#35385518

aroquev/src/main/webapp/script.js
```diff
@@ -346,3 +346,86 @@ function deleteAllComments() {
...
+  const loginInfo = await response.json();
+  return loginInfo;
+}
+
+async function buildWriteCommentsBox() {
+  const writeCommentsBox = document.getElementById('js-write-comment-box');
+  const loginInfo = await checkLoggedIn();
+  if (loginInfo.isLoggedIn) {  
+    const commentForm = document.createElement('form');
+    commentForm.action = 'javascript:;';
```
brettapeters, line 366: This is unnecessary.

aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -59,20 +62,22 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
 
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    JsonObject jsonObj = new Gson().fromJson(request.getReader(), JsonObject.class);
+    String comment = jsonObj.get("commentBody").getAsString();
+
+    UserService userService = UserServiceFactory.getUserService();
+    String username = userService.getCurrentUser().getEmail();
```
brettapeters, line 69: I think most people would not want their email to be used as their username.

aroquev/src/main/webapp/script.js
```diff
@@ -346,3 +346,86 @@ function deleteAllComments() {
...
+  getServletComments();
+}
+
+async function checkLoggedIn() {
+  const response = await fetch('/login-status');
+  const loginInfo = await response.json();
+  return loginInfo;
+}
+
+async function buildWriteCommentsBox() {
```
brettapeters, line 361: This function is getting very long. Consider breaking it into smaller functions with distinct responsibilities.

aroquev/src/main/webapp/script.js
aroquev00: True, done.

aroquev/src/main/webapp/script.js
aroquev00: Cool! Done.

aroquev/src/main/webapp/script.js
aroquev00: I tried to break it apart into two other functions that build the comment box depending on if the user is logged in or not.

aroquev/src/main/java/com/google/sps/servlets/DataServlet.java
aroquev00: Yep, I plan on implementing the 'nickname' feature next to finish this login functionality.

PR #57, opened 2020-07-08
Log in and Log out in comment section

pamelalozano16:
#38  PR Type

- [ ] Bugfix
- [x] Feature

Issue Number: N/A

## What is the new behavior?

-Set nickname (to not show email info)
-Log in user
-Log out user
-Post comment and save author
-Display author comments

## Other information
Screenshot 2020-07-08 at 16 45 22 - Display 2
Screenshot 2020-07-08 at 16 44 39 - Display 2
Screenshot 2020-07-08 at 16 44 10 - Display 2

### DIFF
--- a/pamelalozano/src/main/java/com/google/sps/servlets/Auth.java
+++ b/pamelalozano/src/main/java/com/google/sps/servlets/Auth.java
@@ -45,10 +45,10 @@ public class Auth {
   public static class Builder {
     private boolean isLoggedIn=false;
     private boolean isAdmin=false;
-    private String logoutUrl=null;
-    private String loginUrl=null;
-    private String activeUser=null;
-    private String nickname=null;
+    private String logoutUrl;
+    private String loginUrl;
+    private String activeUser;
+    private String nickname;
 
     public Builder setLoginUrl(String loginUrl) {
       this.loginUrl = loginUrl;
--- a/pamelalozano/src/main/java/com/google/sps/servlets/Auth.java
+++ b/pamelalozano/src/main/java/com/google/sps/servlets/Auth.java
@@ -12,6 +12,7 @@ public class Auth {
   private final String logoutUrl;
   private final String loginUrl;
   private final String activeUser;
+  private final String nickname;
  
   public String getLoginUrl() {
       return this.loginUrl;
@@ -28,6 +29,9 @@ public class Auth {
   public Boolean isAdministrator() {
       return this.isAdmin;
   }
+  public String getNickname() {
+      return this.nickname;
+  }
 
   private Auth(Builder builder) {
     isLoggedIn = builder.isLoggedIn;
@@ -35,6 +39,7 @@ public class Auth {
     activeUser = builder.activeUser;
     loginUrl = builder.loginUrl;
     logoutUrl = builder.logoutUrl;
+    nickname = builder.nickname;
   }
 
   public static class Builder {
@@ -43,6 +48,7 @@ public class Auth {
     private String logoutUrl=null;
     private String loginUrl=null;
     private String activeUser=null;
+    private String nickname=null;
 
     public Builder setLoginUrl(String loginUrl) {
       this.loginUrl = loginUrl;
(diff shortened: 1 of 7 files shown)


pamelalozano/src/main/java/com/google/sps/servlets/Auth.java
```diff
@@ -43,6 +48,7 @@ private Auth(Builder builder) {
     private String logoutUrl=null;
     private String loginUrl=null;
     private String activeUser=null;
+    private String nickname=null;
```
brettapeters, line 51: These primitive data types will have these same values by default.

https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html

pamelalozano/src/main/java/com/google/sps/servlets/NicknameServlet.java
```diff
@@ -0,0 +1,93 @@
...
+
+    response.sendRedirect("/");
+  }
+
+  /**
+   * Returns the nickname of the user with id, or empty String if the user has not set a nickname.
+   */
+  private String getUserNickname(String id) {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Query query =
```
brettapeters, line 82: I think you can create a UserInfo Key directly like `Key userInfoKey = KeyFactory.createKey("UserInfo", id)` and do `Entity entity = datastore.get(key)`. 

I don't think it really makes much of a difference in this case since "id" is indexed. Let me know if this works if you try it out!

pamelalozano/src/main/webapp/script.js
```diff
@@ -20,6 +20,49 @@ document.getElementById("projects").addEventListener("click", function (event) {
 
 let cursor = "";
 
+async function initialLoads(){
+    await loadComments();
```
brettapeters, line 24: Can this be done in parallel?

pamelalozano/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -161,7 +161,7 @@ private Comment getComment(HttpServletRequest request) {
     String msg = request.getParameter("msg");
     String author=null;
     if (userService.isUserLoggedIn()) {
-        author = userService.getCurrentUser().getEmail();
+        author = getUserNickname(userService.getCurrentUser().getUserId());
```
brettapeters, line 164: What if a user wants to change their nickname later? Any old comments would still show their old nickname.

pamelalozano/src/main/webapp/script.js
seanjohnite: +1

pamelalozano/src/main/java/com/google/sps/servlets/DataServlet.java
pamelalozano16: Yeah, there are some things I have thought about but I don't have if there's time to implement, like nickname changes and repeated nicknames. I copied the whole nickname's part from the example in the walkthrough: 
https://github.com/pamelalozano16/STEP/tree/master/walkthroughs/week-4-libraries/authentication/examples/user-nicknames/src/main/java/com/google/sps/servlets

I'll try to figure this out by the end of the week if there's any time left, I think I need to start to implement another library.

pamelalozano/src/main/webapp/script.js
pamelalozano16: Upps yeah, done!

pamelalozano/src/main/java/com/google/sps/servlets/NicknameServlet.java
pamelalozano16: Okay it took me a little long to understand what you meant and it kept throwing an error but I did a little research and the function know looks like this and I tested it it works fine:
private String getUserNickname(String id) {
    Entity entity = null;
   
try {
        Key userInfoKey = KeyFactory.createKey("UserInfo", id);
        entity = datastore.get(userInfoKey);
    } catch(Exception e){
        System.out.println(e);
    }
   
 if (entity == null) {
      return null;
    }
    
   String nickname = (String) entity.getProperty("nickname");
    return nickname;
  }

brettapeters: Nice!

PR #58, opened 2020-07-09
Login nicknames

aroquev00:
I really believe that this picture represents this PR 🤣  It took me waaaaaaay more time that I thought it would, but it works... I think...
image

Jokes aside, the app now allows the user to identify himself by a nickname (that he can later change). All his comments (also the past ones) will be displayed with the nickname he chooses. 
The Java backend is kinda a mess because I did not want to bind the nickname to the Comment entities, and although I will probably have to work on it, I think not binding the nickname is better to provide flexibility when changing it after.
As usual, link to the deployment (nightly release hehe): https://aroquev-step-2020.uc.r.appspot.com/


aroquev/src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -33,19 +38,70 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     
     boolean isLoggedIn = userService.isUserLoggedIn();
     String url;
+    String nickname = "";
 
     if (isLoggedIn) {
       url = userService.createLogoutURL("/");
+      nickname = getUserNickname(userService.getCurrentUser().getUserId());
+      if (nickname == "") {
```
brettapeters, line 46: `nickname.isEmpty()`

aroquev/src/main/java/com/google/sps/data/Comment.java
```diff
@@ -60,4 +69,21 @@ public Entity toEntity() {
...
+   * Returns the nickname of the user with id, or empty String if the user has not set a nickname.
+   */
+  private String getUserNickname(String id) {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Query query =
+        new Query("UserInfo")
+            .setFilter(new Query.FilterPredicate("id", Query.FilterOperator.EQUAL, id));
+    PreparedQuery results = datastore.prepare(query);
+    Entity entity = results.asSingleEntity();
+    if (entity == null) {
```
brettapeters, line 82: FYI you could return an `Optional<String>` from this method to better communicate that a username could be empty. 

https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html

```java
if (entity == null) {
  return Optional.empty();
}
// ...
return Optional.of(nickname);
```

```java
Optional<String> nickname = getUserNickname(...);
if (!nickname.isPresent()) {...}
else {
  System.out.println(nickname.get());
}
```

aroquev00: Nice, would it be good that the attribute of Comment is also an optional, or that one should remain as a String?

brettapeters: You'd probably want to change that to
```java
this.nickname = getUserNickname(username).orElse("anonymous");
```

seanjohnite:
let us know when to merge!

aroquev/src/main/java/com/google/sps/data/Comment.java
aroquev00: Done.

aroquev/src/main/java/com/google/sps/servlets/LoginServlet.java
aroquev00: Done.

aroquev00:
I think its ready!

aroquev00:
Don't know if that works lol but it looks cool when you guys do it.

PR #60, opened 2020-07-09
Add Mexico map

aroquev00:
Added a Mexico Map where I plan to include some suggested destinations (routes as well?) when traveling here. Page is getting too long so also planning on splitting it up.
image

### DIFF
--- a/aroquev/src/main/webapp/script.js
+++ b/aroquev/src/main/webapp/script.js
@@ -330,6 +330,7 @@ function deleteAllComments() {
 }
 
 function onDOMLoad() {
+  createMap();
   buildWriteCommentsBox();
   getServletComments();
 }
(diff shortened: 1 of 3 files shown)


aroquev/src/main/webapp/script.js
```diff
@@ -396,3 +397,10 @@ function submitComment(commentForm) {
     commentForm.reset();
   });
 }
+
+/** Creates a map and adds it to the page. */
+function createMap() {
+  const map = new google.maps.Map(
+      document.getElementById('map'),
+      {center: {lat: 23.167, lng: -101.842}, zoom: 5});
```
brettapeters, line 405: Consider extracting the coordinates object and naming it something descriptive.

PR #62, opened 2020-07-10
Select indicator

pamelalozano16:
#38  PR Type

- [ ] Bugfix
- [ ] Feature

## What is the new behavior?

The user can now select what indicator he wants to see and according to the last update of that indicator the API is called with that year.

## Other information
Screenshot 2020-07-10 at 13 23 33 - Display 2

### DIFF
--- a/pamelalozano/src/main/webapp/maps.js
+++ b/pamelalozano/src/main/webapp/maps.js
@@ -45,7 +45,7 @@ async function createMapChart(indicatorCode, indicatorName, year){
   });
     
       google.charts.setOnLoadCallback(drawRegionsMap);
-    console.log(indicatorCode+" "+indicatorName)
+
       async function drawRegionsMap() {
 
         //ToDo: The indicator code will be the name of the selected indicator in the input element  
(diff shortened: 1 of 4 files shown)


pamelalozano/src/main/webapp/maps.js
```diff
@@ -13,7 +13,31 @@
...
+    document.getElementById("map").innerHTML="";
+    let  form = document.getElementById("selectIndicator");
+    let selected = form.options[form.selectedIndex];
+    let year = checkIndicatorYear(selected.value);
+    document.getElementById("indicatorTitle").innerText=selected.text+" - "+year;
+    createMapChart(selected.value, selected.text, year);
+})
+
+/* Returns the year with the last available information of each indicator*/
+function checkIndicatorYear(indicator) {
```
brettapeters, line 27: A map might be a better than a switch statement for this lookup.

```js
const indicatorYear = {
  'ER.PTD.TOTL.ZS': 2018,
  ...
};
```

```js
let year = indicatorYear[selected.value];
```

pamelalozano16: Okay thanks! Done :)

PR #70, opened 2020-07-14
Add TDD files

aroquev00:
Just add the TDD project files.
I moved the portfolio into its folder and kept the TDD in another folder.


seanjohnite:
ooh can you fix the build? i think you probably just have to change the clouldbuild file

aroquev00:
Mmm like the pom.xml?

aroquev00:
I think it fails because it does not pass the tests yet...

aroquev00:
No wait that is not it

aroquev/.gitignore
```diff
@@ -1 +1 @@
-target/*
+*target/*
```
brettapeters, line 1: You can put a .gitignore in a subdirectory and it will apply to that dir & its subdirectories only.

PR #74, opened 2020-07-16
Passes optional attendees tests and challenge

aroquev00:
The code now passes all extra tests and some others that I added, I think it also solves the extra challenge.


brettapeters:
What's going on with the diff here?

aroquev00:
Went bananas, sorry, just fixed it.

PR #79, opened 2020-07-22
Add Servlet Context listener and Counter and Realtime DAO

aroquev00:
In terms of functionality this does the same thing. However, the app now has DAOs which allow to separate the code logic from the connection to the Realtime Database. RealtimeDao is the implementation of CounterDao and in the servlet it is referenced to retrieve and update the values of the counter in the Realtime DB. I added a ContextListener which is initiated when the app is first run and all servlets can access it (this is where the Firestore DAO is). DataServlet now updates the counter using this DAO.


src/main/java/com/google/sps/daos/RealtimeDao.java
```diff
@@ -0,0 +1,74 @@
...
+
+    } catch(IOException ie) {
+      ie.printStackTrace();
+    }
+    
+  }
+
+  @Override
+  public Counter getCounter() {
+    return counter;
```
brettapeters, line 61: Should this be reading the current counter value from the db?

src/main/java/com/google/sps/util/QuizzyContextListener.java
```diff
@@ -0,0 +1,26 @@
...
+  @Override
+  public void contextDestroyed(javax.servlet.ServletContextEvent event) {
+  }
+
+  @Override
+  public void contextInitialized(ServletContextEvent event) {
+    // This function is called when the application starts and will safely set a few required
+    // context attributes such as the RealtimeDao
+
+    RealtimeDao dao = (RealtimeDao) event.getServletContext().getAttribute("dao");
```
brettapeters, line 20: You can just use the interface type (CounterDao) instead of the more precise implementation type (RealtimeDao)

src/main/java/com/google/sps/daos/RealtimeDao.java
```diff
@@ -0,0 +1,74 @@
...
+public class RealtimeDao implements CounterDao {
+  
+  private Counter counter;
+
+  private DatabaseReference realtimeDb;
+
+  public RealtimeDao() {
+
+    try {
+      FirebaseOptions options = new FirebaseOptions.Builder()
```
brettapeters, line 25: You might want to initialize Firebase in the ServletContextListener and then pass the app instance to this constructor. That way, any other DAOs in the future can reuse it.

I also think it will give you an error if you try to initialize Firebase more than once.

src/main/java/com/google/sps/daos/RealtimeDao.java
aroquev00: Okay so this is a design question I have: the way it is now, because of the listener added at the top of the file, counter gets updated every time there is a change, so the idea is that returning counter should mean that that counter is the most up to date counter. However, this is not very good design right? It should query the DB every time it is asked to do so right? Also, we probably would be wasting CPU on updating the variable counter even though it is not consulted by the user?

src/main/java/com/google/sps/util/QuizzyContextListener.java
aroquev00: Done.

src/main/java/com/google/sps/daos/RealtimeDao.java
brettapeters: 1) It looks like the `counter` in memory here is never updated. So the code in `doGet()` will always result in writing `1`

2) You should be updating the counter value inside a [transaction](https://firebase.google.com/docs/database/admin/save-data#section-transactions). In the (probable) case that two requests to update the counter are being handled at the same time, they will both write the same value to the db unless you do the counter read/write inside a transaction.

aroquev00: Regarding point 1, doesn't it get updated in line 43?

Agree with point #2, will do!

brettapeters: Oh I see. You still have the same problem with concurrent updates using stale data doing it that way.

aroquev00: Gotcha, aside from that, do you think that there would be a point on keeping the `counter` object in memory or should I just query it and send it to the user every time it is requested?

brettapeters: Yeah it should be read from the db every time. The db is the single source of truth.

src/main/java/com/google/sps/daos/RealtimeDao.java
aroquev00: Done.

aroquev00:
I changed it so that now the counter is updated through a transaction so that there is no missed information. The counter is not retrieved in the server because the client will be listening to it.

PR #82, opened 2020-07-24
Connect back front realtime

aroquev00:
So this is the thing that should work, but doesn't... Everything is hooked up, running locally the client listens to the Realtime DB and does update the counter when it is updated through the servlet via a POST request to servlet /incrase. The thing is that when deployed, the servlet does not connect to the realtimeDB we can see in the console (https://quizzy-step-2020.firebaseio.com/). How to test it? Click on the Vote button `n` times. You will see that the counter is not updated. Then make a get request by going to https://quizzy-step-2020.uc.r.appspot.com/store. By doing this, the servlet queries the Realtime DB counter value (which has you magical `n` number) and stores it in Firestore. You should be able to see that here https://firebase.corp.google.com/project/quizzy-step-2020/database/firestore/data~2Fcounter~2FNiqBOLKdR7hXBWk4Awa5 as a new document in the `counter` collection. If this was not clear, ping me so that I can explain over GVC!
Here is the deployed one as usual: https://quizzy-step-2020.uc.r.appspot.com/


aroquev00:
Okay I think it's ready now and hooked up with Firestore: https://quizzy-step-2020.uc.r.appspot.com/

src/main/webapp/index.html
```diff
@@ -18,11 +18,15 @@
...
+  <link rel="icon" 
+      type="image/png" 
+      href="images/noogler-hat.png">
+
   <!-- Disable tap highlight on IE -->
   <meta name="msapplication-tap-highlight" content="no">
 
   <!-- Web Application Manifest -->
-  <link rel="manifest" href="manifest.json">
+  <!-- <link rel="manifest" href="manifest.json"> -->
```
seanjohnite, line 29: maybe just delete this guy?

aroquev00: Done.

src/main/webapp/index.html
```diff
@@ -105,6 +106,7 @@ <h1 id="counter"></h1>
   <script src="https://www.gstatic.com/firebasejs/7.16.1/firebase-performance.js"></script>
   <script src="https://www.gstatic.com/firebasejs/7.16.1/firebase-auth.js"></script>
   <script src="https://www.gstatic.com/firebasejs/7.16.1/firebase-firestore.js"></script>
+  <script src="https://www.gstatic.com/firebasejs/7.16.1/firebase-database.js"></script>
```
thefredo1000, line 109: I think this is not required anymore.

aroquev00: Yeah you're right, thanks!

brettapeters:
I wonder how we can avoid the long delay when the app hasn't been used in a while?? Probably not worth worrying about too much, but I am curious why it behaves this way.

PR #85, opened 2020-07-27
Create room

pamelalozano16:
#84 PR Type

- [ ] Bugfix
- [x] Feature

## What is the new behavior?

The "/newRoom" creates a new room for a specific game (right now its hard coded), saves that game id and saves the liveRoom id in that game document. When you get "/startGame" the isActive in live room becomes true, in that moment it's going to get the questions and answers and if you go to "/joinRoom" it joins to the actual room, the one that it was just created (for testing reasons). Later it will join whatever room is passed.


src/main/java/com/google/sps/daos/DatabaseRoomDao.java
```diff
@@ -0,0 +1,75 @@
...
+
+    newRoomRef.set(data);
+
+    //Room document is now created
+    this.actualRoom = newRoomRef.getId();
+
+   //Updates game document and adds liveRoom id
+   Map<String, Object> update = new HashMap<>();
+   update.put("liveRoom", this.actualRoom);
+   ApiFuture<WriteResult> writeResult = firestoreDb.collection("games").document(this.gameId).set(update, SetOptions.merge());
```
aroquev00, line 50: Just gotta figure out how we are going to handle the multiple games using same questions thing.

src/main/java/com/google/sps/daos/DatabaseRoomDao.java
```diff
@@ -0,0 +1,75 @@
...
+    firestoreDb.collection("liveRooms").document(roomId).collection("members").add(data);
+  }
+
+ @Override
+ public void startGame(){
+    Map<String, Object> update = new HashMap<>();
+    update.put("isActive", true);
+    ApiFuture<WriteResult> writeResult = firestoreDb.collection("liveRooms").document(this.actualRoom).set(update, SetOptions.merge());
+ }
+}
```
aroquev00, line 75: add newline?

src/main/java/com/google/sps/daos/RoomDao.java
```diff
@@ -0,0 +1,8 @@
+package com.google.sps.daos;
+
+public interface RoomDao {
+  public String createNewRoom(String gameId);
+  public String getActualRoom();
+  public void startGame();
+  public void joinRoom(String roomId, String uId);
+}
```
aroquev00, line 8: newline?

src/main/java/com/google/sps/servlets/CreateRoomServlet.java
```diff
@@ -0,0 +1,39 @@
...
+    
+    String json = new Gson().toJson(jsonObj);
+    response.getWriter().println(json);
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Generate room key
+    RoomDao dao = (RoomDao) this.getServletContext().getAttribute("roomDao");
+    dao.createNewRoom("5GoXkueDimFjk1wEBLhV");
```
thefredo1000, line 35: Isn't `RoomDao` an interface?
```suggestion
    DatabaseRoomDao dao = (RoomDao) this.getServletContext().getAttribute("roomDao");
    dao.createNewRoom("5GoXkueDimFjk1wEBLhV");
```

src/main/java/com/google/sps/servlets/StartGame.java
```diff
@@ -0,0 +1,34 @@
...
+    dao.startGame();
+    //Todo: get room's game and get questions
+
+    response.getWriter().println(dao.getActualRoom());
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+  }
```
thefredo1000, line 32: Delete?

src/main/java/com/google/sps/servlets/CreateRoomServlet.java
aroquev00: Yes it is, but @brettapeters told us that this is the best way to have the DAO.

brettapeters: Right. Same as you would write `List<String> names = new ArrayList<>();` `List` is an interface and `ArrayList` is an implementation of a `List`. This servlet doesn't need to know which implementation of RoomDao it's using.

brettapeters: Kind of sort of related to the "I" in SOLID. https://en.wikipedia.org/wiki/SOLID

src/main/java/com/google/sps/daos/DatabaseRoomDao.java
```diff
@@ -0,0 +1,75 @@
...
+
+    return this.actualRoom;
+  }
+ 
+ @Override
+  public String getActualRoom() {
+      return this.actualRoom;
+  }
+
+  @Override
```
brettapeters, line 61: I feel like there should be a separation between `RoomDao` and a `Room` where `RoomDao` just deals with CRUD operations on a `Room`, and `Room` is just a POJO.

`joinRoom` and `startGame` seem like they could both be represented as "updates" to a `Room` entity.

src/main/java/com/google/sps/servlets/StartGame.java
aroquev00: I think she has it in doGet for now, but later it will be in doPost because you send a POST request to start a new game right?

src/main/java/com/google/sps/daos/DatabaseRoomDao.java
aroquev00: @brettapeters so if the purpose of `joinRoom` is to just add a member to the DB's list of members for a given room so that the client can listen to it how do you think we should separate? Maybe in the Servlet creating a `Room` instance that represents the actual state of the room in the DB and update it there and then save the updated `Room` instance to the DB?

src/main/java/com/google/sps/servlets/StartGame.java
pamelalozano16: Yes exactly I think it's going to be useful later but I can delete it for now

src/main/java/com/google/sps/daos/DatabaseRoomDao.java
pamelalozano16: Yeah so I think Brett means creating a Room entity and then using the Dao to modify it, like the questions example in the design doc, and once modified local pass the update of the Room entity to the db ? @brettapeters

src/main/java/com/google/sps/daos/DatabaseRoomDao.java
pamelalozano16: Yeahh I think as we move on we'll figure out that part

src/main/java/com/google/sps/daos/DatabaseRoomDao.java
```diff
@@ -0,0 +1,75 @@
...
+import com.google.cloud.firestore.DocumentSnapshot;
+import com.google.cloud.firestore.WriteResult;
+import com.google.firestore.v1.StructuredQuery.CollectionSelector;
+
+public class DatabaseRoomDao implements RoomDao {
+
+  private DatabaseReference roomsDBReference;
+  private Firestore firestoreDb;
+  private String actualRoom;
+  private String gameId;
```
brettapeters, line 22: Since `gameId` is part of a `Room`, does the servlet need to store a reference to it?

src/main/java/com/google/sps/daos/DatabaseRoomDao.java
```diff
@@ -0,0 +1,75 @@
...
+
+  public DatabaseRoomDao(Firestore firestoreDb) {
+    this.firestoreDb = firestoreDb;
+  }
+  
+  @Override
+  public String createNewRoom(String gameId) {
```
brettapeters, line 29: Maybe this should take a `Room` instance as its argument. `public Room create(Room room);`

src/main/java/com/google/sps/servlets/StartGame.java
```diff
@@ -0,0 +1,29 @@
...
+
+@WebServlet("/startGame")
+public class StartGame extends HttpServlet {
+  
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException { 
+
+    // Activate room
+    RoomDao dao = (RoomDao) this.getServletContext().getAttribute("roomDao");
+    dao.startGame();
```
seanjohnite, line 23: I think Brett is right here. If we want to do something like `thing.startGame()`, that thing should probably not be the RoomDao. Maybe RoomHelper/GameInstanceHelper or something?

src/main/java/com/google/sps/daos/DatabaseRoomDao.java
seanjohnite: Right.. Another way to ask the same question: what happens to the `gameId` and `actualRoom` properties if more than one room exists where games are being played?

pamelalozano16: Each room has its own game? So if you want to play a game it generates its own room, you can't play several games in one room because it's the 'game instance'. But I till delete it from the dao it's not necessary there.

src/main/java/com/google/sps/daos/DatabaseRoomDao.java
pamelalozano16: Oh yeah true let me do it

src/main/java/com/google/sps/servlets/StartGame.java
pamelalozano16: Yeah it is better I will try to change everything tomorrow :)

aroquev00:
Yeah I think we should refactor the names so that it's ´GameInstance´

src/main/java/com/google/sps/daos/DatabaseRoomDao.java
pamelalozano16: Can you look the changes ?? I created the room inside the createNewRoom method, instead of passing it as an argument, is it the same or is it better this way?

src/main/java/com/google/sps/servlets/StartGame.java
pamelalozano16: I'll do this after this PR merge :)

src/main/java/com/google/sps/daos/DatabaseRoomDao.java
```diff
@@ -0,0 +1,75 @@
...
+  public String getActualRoom() {
+      return this.actualRoom;
+  }
+
+  @Override
+  public void joinRoom(String roomId, String uId) {
+    Map<String, Object> data = new HashMap<>();
+    data.put("UID", uId);
+
+    firestoreDb.collection("liveRooms").document(roomId).collection("members").add(data);
```
seanjohnite: This is adding to the members collection rather than the members list, right?

src/main/java/com/google/sps/daos/DatabaseRoomDao.java
seanjohnite: Yes, but this is the implementation of the interface that handles interacting with the database, correct? So this is assuming that only one room/game exists at a time in the whole world, right?

src/main/java/com/google/sps/daos/DatabaseRoomDao.java
```diff
@@ -0,0 +1,75 @@
...
+import com.google.firestore.v1.StructuredQuery.CollectionSelector;
+
+import com.google.sps.data.Member;
+import com.google.sps.data.Room;
+
+public class DatabaseRoomDao implements RoomDao {
+
+  private DatabaseReference roomsDBReference;
+  private Firestore firestoreDb;
+  private String actualRoom;
```
seanjohnite, line 28: This is not used anymore correct? Also `roomsDBReference`

src/main/java/com/google/sps/daos/DatabaseRoomDao.java
pamelalozano16: Yess I changed that in the next commit

src/main/java/com/google/sps/daos/DatabaseRoomDao.java
pamelalozano16: Yess you are right, done!

src/main/java/com/google/sps/daos/DatabaseRoomDao.java
brettapeters: It can work this way. My reasoning for passing a `Room` instance here is to keep the method signature more flexible. `Room` can have fields added/removed in the future but the DAO wouldn't have to change. If gameId is the only bit of info ever needed to construct a new game, then this will work.

PR #93, opened 2020-07-30
New game instance attributes

pamelalozano16:
##PR Type

- [ ] Bugfix
- [x] Feature
- [x] Code style update (formatting, local variables)

## What is the new behavior?

Basically what we want is a servlet that contains the game instance information and the actual question information, to get question and answers and in the front end, I'm thinking of using post to start the game and to get next question, I think the name of the servlet will change to /actualGame or something more descriptive, the parameter "room" will contain the game instance number.

## Other information
The json in /actualGame would be:
`{
    gameInstance: {...},
    question: {...}
}`
Based on the currentQuestion id in game instance


src/main/java/com/google/sps/servlets/StartGameInstanceServlet.java
```diff
@@ -29,17 +29,22 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
     GameInstance newRoom = dao.getRoom(roomId);
-    // Activate room
-    if(newRoom != null){
-        newRoom.setIsActive(true);
-        dao.updateRoom(newRoom);
-        response.getWriter().println("Game started");
-    } else {
+
+    if(newRoom == null){
         response.setStatus(404);
```
thefredo1000, line 34: We should create a file to handle 404

src/main/java/com/google/sps/data/GameInstance.java
```diff
@@ -9,6 +9,8 @@
   private boolean isActive;
   private String creator;
   private String gameId;
+  private String currentQuestion;
```
aroquev00, line 12: Nice, we're getting to how to control the state of the game. If we store the `currentQuestion` as a String, this would mean copying from a `Game` to the `GameInstance` right? But we would also need to copy the answers for it to be independent of the `Game`? The other thing I was thinking about is just storing `currentQuestion` as an `int` that just signals which `Question` to retrieve from `Game` in the front end.

PR #99, opened 2020-08-03
Demo tomorrow in live tv sos

pamelalozano16:
PR Type

- [ ] Bugfix
- [x] Feature

## What is the new behavior?

The teacher control is now from the front to the gameInstance document, it changes the question id and then queries that question to get data in the front end.

## Other information
Screenshot 2020-08-03 at 18 12 20 - Display 2
Screenshot 2020-08-03 at 18 11 46 - Display 2

### DIFF
--- a/src/main/java/com/google/sps/servlets/StartGameInstanceServlet.java
+++ b/src/main/java/com/google/sps/servlets/StartGameInstanceServlet.java
@@ -2,6 +2,7 @@ package com.google.sps.servlets;
 
 import com.google.gson.Gson;
 import com.google.gson.JsonObject;
+import java.util.List; 
 
 import java.io.IOException;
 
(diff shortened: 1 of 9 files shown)


src/main/java/com/google/sps/servlets/StartGameInstanceServlet.java
```diff
@@ -29,15 +47,43 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
     }
-    
-    // Activate room
-    newRoom.setIsActive(true);
+    if (action.equals("end")) {
+        newRoom.setIsActive(false);
+        newRoom.setCurrentQuestion(null);
+    }
+    if (action.equals("start")) {
+        ApiFuture<QuerySnapshot> future = db.collection("games").document(newRoom.getGameId()).collection("questions").get();
```
brettapeters, line 61: Are questions always returned in the proper sorted order?

src/main/java/com/google/sps/servlets/StartGameInstanceServlet.java
```diff
@@ -2,24 +2,42 @@
...
 
+import java.util.List; 
+import java.util.ArrayList; 
+
 @WebServlet("/startGameInstance")
 public class StartGameInstanceServlet extends HttpServlet {
-  
+
+    private List<QueryDocumentSnapshot> questions = new ArrayList<>();
+    private int indexQuestions;
```
brettapeters, line 33: I think this index should be a property of GameInstance and not stored in the servlet's state. If the server shuts down and/or two requests are handled by different server instances this index would be incorrect.

brettapeters: And furthermore, it should be updated in a transaction to ensure concurrent requests do not read a stale value.

src/main/java/com/google/sps/servlets/StartGameInstanceServlet.java
```diff
@@ -2,24 +2,42 @@
...
+
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException { 
   // Generate room key
     String roomId = request.getParameter("gameInstance");
+    String action = request.getParameter("action");
```
brettapeters, line 39: Validate?

src/main/java/com/google/sps/servlets/StartGameInstanceServlet.java
pamelalozano16: > TODO in a followup PR: Separate each action into its own servlet and use POST requests for state changes.

Yeah, did this a little hurried because of the demo presentation but will do!

pamelalozano16:
> TODO in a followup PR: Separate each action into its own servlet and use POST requests for state changes.

Yeah, did this a little hurried because of the demo presentation but will do!

src/main/java/com/google/sps/servlets/StartGameInstanceServlet.java
pamelalozano16: Okay you are right

src/main/java/com/google/sps/servlets/StartGameInstanceServlet.java
pamelalozano16: I don't know but I think I can sort them by creation date? But I think rodrigo will post the questions all at once, not sure how to do it, do you have an idea?

brettapeters: So according to the docs a query returns documents in ascending order by document ID. So if the doc IDs are set up that way it should work as you have it now. But if not, there could be a field for index - then in the query add `.orderBy("index")`. Or (maybe a little less friendly approach) store `nextQuestionId` and maybe `previousQuestionId` in a question doc to form a linked list.

PR #104, opened 2020-08-05
Questions to linked list and refactoring servlets

pamelalozano16:
## PR Type

- [ ] Bugfix
- [ ] Feature

## What is the new behavior?

So now there is a servlet for every action, and it's a post request, not a get. Each game will contain a field called headQuestion, which is the id for the head of the list, in start game instance it will use that head to start going through the list. /nextQuestion uses nextQuestion field in each question and /previousQuestion uses previous question field.
All the servlets take the gameInstance ID as a parameter

### DIFF
--- a/src/main/java/com/google/sps/servlets/EndGameInstanceServlet.java
+++ b/src/main/java/com/google/sps/servlets/EndGameInstanceServlet.java
@@ -10,15 +10,6 @@ import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
-import com.google.cloud.firestore.Firestore;
-import com.google.cloud.firestore.FirestoreOptions;
-import com.google.cloud.firestore.DocumentReference;
-import com.google.cloud.firestore.CollectionReference;
-import com.google.cloud.firestore.QueryDocumentSnapshot;
-import com.google.cloud.firestore.QuerySnapshot;
-import com.google.cloud.firestore.Query;
-import com.google.api.core.ApiFuture;
-import com.google.cloud.firestore.DocumentSnapshot;
 
 import com.google.sps.daos.GameInstanceDao;
 import com.google.sps.data.GameInstance;
@@ -34,7 +25,6 @@ public class EndGameInstanceServlet extends HttpServlet {
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException { 
   // Generate room key
     String roomId = request.getParameter("gameInstance");
-    Firestore db = (Firestore) this.getServletContext().getAttribute("firestoreDb");
 
     if(roomId==null || roomId.isEmpty()){
         response.setStatus(500);
@@ -55,11 +45,6 @@ public class EndGameInstanceServlet extends HttpServlet {
     newRoom.setCurrentQuestion(null);
 
     dao.updateGameInstance(newRoom);
-
-    //Return object
-    Gson gson = new Gson();
-    String json = gson.toJson(newRoom);
-    response.getWriter().println(json);
 
 
   }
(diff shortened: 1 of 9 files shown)


src/main/webapp/scripts/controlGameInstance.js
```diff
@@ -155,17 +155,18 @@ function initUIButtons(gameInstanceId) {
...
   });
   previousQuestionButton.addEventListener('click', () => {
-      fetch('/startGameInstance?action=previous&gameInstance='+activeGameInstanceId);
+      fetch('/previousQuestion?gameInstance='+activeGameInstanceId, { method: 'POST' });
   });
   endGameInstanceButton.addEventListener('click', () => {
-      fetch('/startGameInstance?action=end&gameInstance='+activeGameInstanceId);
-      window.location.href = "/teacher/controlGameInstance.html";
+      fetch('/endGameInstance?gameInstance='+activeGameInstanceId, { method: 'POST' }).then(()=>{
+          window.location.href = "/teacher/controlGameInstance.html";
```
aroquev00, line 168: So when the teacher ends the game right now it just resets the page as if the game had never happened right?

src/main/java/com/google/sps/servlets/EndGameInstanceServlet.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.io.IOException;
+
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import com.google.cloud.firestore.Firestore;
+import com.google.cloud.firestore.FirestoreOptions;
+import com.google.cloud.firestore.DocumentReference;
```
aroquev00, line 15: nit: remove unused Firebase imports?

src/main/java/com/google/sps/servlets/EndGameInstanceServlet.java
```diff
@@ -0,0 +1,67 @@
...
+
+    newRoom.setIsActive(false);
+    newRoom.setCurrentQuestion(null);
+
+    dao.updateGameInstance(newRoom);
+
+    //Return object
+    Gson gson = new Gson();
+    String json = gson.toJson(newRoom);
+    response.getWriter().println(json);
```
aroquev00, line 62: Why are we returning the `newRoom` to the client?

src/main/webapp/scripts/controlGameInstance.js
pamelalozano16: Yes exactly, it is as the initial state before the start.

src/main/java/com/google/sps/servlets/EndGameInstanceServlet.java
pamelalozano16: Oh truee thanks

src/main/java/com/google/sps/servlets/EndGameInstanceServlet.java
pamelalozano16: I guess I thought it would be useful later but I can delete this for now.

src/main/java/com/google/sps/servlets/PreviousQuestionServlet.java
```diff
@@ -0,0 +1,63 @@
...
+import com.google.cloud.firestore.DocumentSnapshot;
+
+import com.google.sps.daos.GameInstanceDao;
+import com.google.sps.data.GameInstance;
+
+import java.util.List; 
+import java.util.ArrayList; 
+
+@WebServlet("/previousQuestion")
+public class PreviousQuestionServlet extends HttpServlet {
```
seanjohnite, line 24: Seems like there are pretty similar functionalities in next and previous at least, maybe even similar workflows otherwise. Can you extract some of this logic to a helper class?

PR #107, opened 2020-08-06
Refactor control game instance

### LINKED ISSUE #96: Change Teachers Admin Panel to use Promises
Right now it's chained functions so it's really hard to read... I must change it to work with promises

aroquev00:
I just refactored the `controlGameInstance.js` file to include `async` and `await` to improve readability and make functions have unique responsability. Now functions have very specific tasks instead of being chained like previously.
I have one question though, what should I do to handle errors? Should I have the functions that query something from Firestore catch any error in case the document doesn't exist and then return it and catch it in the calling function?

Closes #96


src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
```diff
@@ -74,15 +98,15 @@ function queryActiveGameInstanceDocument(gameInstanceId) {
 }
 
 // Build the UI elements with all the info of the GameInstance
-function buildActiveGameInstanceUI(gameInstance, gameInstanceId) {
+function buildActiveGameInstanceUI(gameInstanceId, gameInstance, game) {
```
brettapeters, line 101: This would benefit from using named parameters so you don't have to worry about argument ordering. The common way people do named parameters in JS is this.

```js
function buildActiveGameInstanceUI({ gameInstanceId, gameInstance, game } = {}) {...}

buildActiveGameInstanceUI({ gameInstanceId: 123, gameInstance: gameInstance, game: game });

// Which can be shortened to this
buildActiveGameInstanceUI({ gameInstanceId: 123, gameInstance, game });
```

brettapeters:
You could throw a new error inside the functions that query Firebase if the doc isn't found and catch inside the calling function as long as you want to handle errors the same way in each situation.

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
```diff
@@ -16,34 +16,50 @@
...
-  firebase.auth().onAuthStateChanged(authStateObserver);
+async function loadControlPanel(user) {
+  // Get the Game Instance's ID in which the user is participating
+  const gameInstanceId = await getActiveGameInstanceId(user);
+
+  // Get the Active Game Instance Object from DB
+  const gameInstance = await queryActiveGameInstance(gameInstanceId);
+
+  // Get the Game the GameInstance is using from DB
+  const game = await queryGameDetails(gameInstance.gameId);
```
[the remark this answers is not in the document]
aroquev00: Sure! But this would also mean three sequential reads in the servlet right? Is it still better to have these reads on the back end?

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
```diff
@@ -16,34 +16,50 @@
...
+function initAuthStateObserver() {
+  firebase.auth().onAuthStateChanged(authStateObserver);
+}
 
 // Triggers when the auth state change for instance when the user signs-in or signs-out.
 function authStateObserver(user) {
   if (user) { // User is signed in!
     // Everything starts working when the User logs in
-    getActiveGameInstanceId(user);
+    //getActiveGameInstanceId(user);
```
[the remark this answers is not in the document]
aroquev00: Done.

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
aroquev00: So nice!! So in the last example I could call `buildActiveGameInstanceUI` with just `gameInstanceId, gameInstance, game` in any order (as long as they're named as the parameters of the function)?

aroquev00: Like this: `buildActiveGameInstanceUI({ game, gameInstanceId, gameInstanceId });`

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
```diff
@@ -16,54 +16,82 @@
...
-  db.collection("users").doc(uid).get().then(function(doc) {
+  return db.collection("users").doc(uid).get().then(function(doc) {
     if (doc.exists) {
       // Get the activeGameInstance's ID in which the User is participating
-      activeGameInstanceId = doc.data().activeGameInstanceId;
-      // Query the info of that GameInstance from the "GameInstance" collections
-      queryActiveGameInstanceDocument(activeGameInstanceId);
+      return doc.data().activeGameInstanceId;
     } else {
+      throw 'User not registered in a GameInstance'
```
aroquev00, line 63: @brettapeters So for example, there is a possibility that a user accesses the "Play game" page without having registered in any gameInstance previously, so in the loading process of the game, this read will fail and doc.data() will be null. In that case I want to stop everything and tell display in the UI something like "go register first and then come back". At which point of the calling functions should I place a `catch` to abort loading the rest of the UI based on the nonexistent result of this query? In `authStateObserver` which is the function that calls `loadControlPanel`, where this query is called and the result used?

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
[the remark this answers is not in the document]
aroquev00: This gets fixed in another PR! So it's done!

PR #108, opened 2020-08-06
Test for question control servlets

pamelalozano16:
PR Type

- [ ] Bugfix
- [x] Feature

## What is the new behavior?

For now it's just the StartGameInstanceServlet test

### DIFF
--- a/src/test/java/com/google/sps/StartGameInstanceServletTest.java
+++ b/src/test/java/com/google/sps/StartGameInstanceServletTest.java
@@ -53,6 +53,7 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.Assert;
 import org.junit.runners.JUnit4;
+import org.hamcrest.CoreMatchers;
 
 import java.util.ArrayList;
 
@@ -66,7 +67,10 @@ public final class StartGameInstanceServletTest {
   private HttpServletResponse response = mock(HttpServletResponse.class);
   private String roomId = "aSpX3cmZa5PB994uEoW2";
   private String gameId = "iBxba1vsaWT1SIqxeonJ";
+  private String questionId = "NWUzaBz7SJEiKvQEwAkt";
   private StartGameInstanceServlet servletUnderTest;
+  private GameInstance newRoom;
+  private StringWriter responseWriter;
 
   @Before
   public void setUp() throws Exception {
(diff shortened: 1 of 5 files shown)


pamelalozano16:
Guys I don't want to merge this branch yet, but I'm doing this pull request so you can review this test I did, I'd like all of your reviews because it's the first one I do and I want to make sure I'm going in the right path. Thank you u are awesome :)

pamelalozano16:
> Guys I don't want to merge this branch yet, but I'm doing this pull request so you can review this test I did, I'd like all of your reviews because it's the first one I do and I want to make sure I'm going in the right path. Thank you u are awesome :)

While creating the test I thought of refactoring the StartGameInstanceServlet by using GameDao to get game, that way I wouldn't need to initialize firestore in the test, but we don't have a getGame in GameDao yet.

src/test/java/com/google/sps/StartGameInstanceServletTest.java
```diff
@@ -0,0 +1,112 @@
...
+          .questions(new ArrayList<Question>())
+          .build();
+
+    when(mockGameDao.getGame(gameId)).thenReturn(newGame);
+
+    //Simulate Room to be updated
+    GameInstance newRoom = new GameInstance(roomId);
+    newRoom.setGameId(gameId);
+    newRoom.setCreator("UfQ5TlrtFtNJnR1ywXb8W7Hrz6y1");
+    newRoom.setIsActive(true);
```
brettapeters, line 99: Start out with "isActive" and "currentQuestion" unset before calling `doPost`. Then you can verify their values before/after the servlet code sets them. You could capture the `GameInstance` argument from the call to `updateGameInstance` and assert that its "isActive" and "currentQuestion" fields are certain values.

https://www.javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/ArgumentCaptor.html

brettapeters:
> > Guys I don't want to merge this branch yet, but I'm doing this pull request so you can review this test I did, I'd like all of your reviews because it's the first one I do and I want to make sure I'm going in the right path. Thank you u are awesome :)
> 
> While creating the test I thought of refactoring the StartGameInstanceServlet by using GameDao to get game, that way I wouldn't need to initialize firestore in the test and it would be much faster, but we don't have a getGame in GameDao yet.

Tests shouldn't import anything Firestore related. All you need are the DAO interfaces to use for mocks.

src/test/java/com/google/sps/StartGameInstanceServletTest.java
pamelalozano16: Done! Thanks brett

PR #109, opened 2020-08-07
createGame front-end restructured

thefredo1000:
## PR Checklist

- [ ] PR has a descriptive title
- [ ] PR contains closely related changes only (should not have many unrelated changes)
- [ ] STEP pod members have been added as reviewers

### Before
Screenshot 2020-08-06 at 6 49 33 PM - Display 3
Before this PR we had a fixed amount of questions to have, the "New Question" button didn't work and everything was written in pure html.

### Now
Screenshot 2020-08-06 at 6 54 28 PM - Display 3
We can add as many questions as we want with the "New Question" button.

Screenshot 2020-08-06 at 6 49 49 PM - Display 3
When we try to add a new question a modal will show up to the user with three options, if we want to create a multiple choice question or a true or false question. 

## What's next?

- Delete questions
- Submit to a servlet
- Make it look nicer


src/main/webapp/scripts/createGame.js
```diff
@@ -20,3 +35,140 @@ function hideAllExcept(ref) {
...
   }
 }
+
+// TODO: Find a better way to tell this function what type of question it is
+function addNewQuestion(isMC) {
+  // Sets the header if idMC (is multiple choice) is true or not
+  var header = isMC ? "Multiple Choice" : "True or False";
+
+  let formElement = document.getElementById("createGameForm");
+  let num = formElement.children.length -2; // This is the index for the next question
```
brettapeters, line 45: This seems like this code could break easily if the HTML structure changes. Is there another way to get this value that is not coupled to the HTML structure?

pamelalozano16:
> The UI looks great! Congrats! I think it's only missing the linked list when you submit the game to the servlet, meanwhile it will not work in the teacher/student game.

Also a nice add would be a variable of how much time the question is available for students to answer in the game (ex 1 min in true or false)  and a variable of what type of question it is, this would be really helpful when doing the front of the game

thefredo1000:
> > The UI looks great! Congrats! I think it's only missing the linked list when you submit the game to the servlet, meanwhile it will not work in the teacher/student game.
> 
> Also a nice add would be a variable of how much time the question is available for students to answer in the game (ex 1 min in true or false) and a variable of what type of question it is, this would be really helpful when doing the front of the game

You can't submit this form right now, this was just UI changes, I'm going to add that on the next PR. The thing you've said about the time it's actually a really good idea and I'll make a commit to add that.

src/main/webapp/scripts/createGame.js
thefredo1000: I can have a global variable that stores the amount of questions.

PR #112, opened 2020-08-10
Verify user in firestore

aroquev00:
I added the functionality that whenever a user signs in our web page, the user is registered in the Firestore in the collection "Users". I only register the Firebase ID that Firebase Auth generates for each user in our app. This will allow us to register in which game instance the User is present and other nice features. As far as I understand this is not PII since the UID in meaningless outside the context of our application, right?

The way it works is that whenever a user signs in, the client sends a request to the back end with a Firebase Auth Token to asynchronously verify if the user is already registered in the "Users" collection. If not, it registers the user.


src/main/java/com/google/sps/daos/UserDao.java
```diff
@@ -0,0 +1,5 @@
+package com.google.sps.daos;
+
+public interface UserDao {
+  public void verifyUserInFirestore(String idToken);
```
brettapeters, line 4: I have seen this type of method usually called `createIfNotExists`

aroquev00: Nice, done!

aroquev00:
If no more comments, this one is also ready to merge into master!

PR #115, opened 2020-08-10
PreviousQuestionServlet and EndGameInstance tests

pamelalozano16:
PR Type

- [ ] Bugfix
- [x] Feature

## What is the new behavior?

Test for previous question servlet. Game dao now contains a getQuestionId function that takes "nextQuestion" or "previousQuestion" as parameters.

### DIFF
--- a/src/main/java/com/google/sps/daos/DatabaseGameDao.java
+++ b/src/main/java/com/google/sps/daos/DatabaseGameDao.java
@@ -83,4 +83,17 @@ public class DatabaseGameDao implements GameDao {
       }
       return newGame;
     }
+
+@Override
+    public String getNextQuestionId(String gameId, String currentQuestionId) {
+      ApiFuture<DocumentSnapshot> future = firestoreDb.collection("games").document(gameId).collection("questions").document(currentQuestionId).get();
+        try {
+            DocumentSnapshot document = future.get();
+            return document.get("nextQuestion").toString();
+        } catch(Exception e) {
+            System.out.println(e);
+            return "";
+        }
+
+    }
 }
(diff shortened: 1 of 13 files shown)


src/main/java/com/google/sps/daos/DatabaseGameDao.java
```diff
@@ -83,4 +83,18 @@ public Game getGame(String id) {
       }
       return newGame;
     }
+
+
+@Override
+    public String getQuestionId(String question, String gameId, String currentQuestionId) {
```
brettapeters, line 89: `nextQuestionId` and `previousQuestionId` could be fields on `Question`. So when you have a reference to the current question, just call `currentQuestion.nextQuestionId()` and update the `GameInstance` with that value.

pamelalozano16: I think that's more less what I'm doing here, I would still have to get the question from the db and `getNextQuestion()` right? 
In here I'm doing the same but I'm just doing `document.get("nextQuestion").toString();`
Correct me if i'm wrong maybe I didn't understood

src/main/java/com/google/sps/daos/DatabaseGameDao.java
```diff
@@ -84,12 +84,14 @@ public Game getGame(String id) {
       return newGame;
     }
 
+
 @Override
-    public String getNextQuestionId(String gameId, String currentQuestionId) {
+    public String getQuestionId(String question, String gameId, String currentQuestionId) {
```
aroquev00, line 89: What is the purpose of the first `question` parameter?

pamelalozano16: Oh it's to define if it is looking for the next question or the previous question, because the functions are almost the same, so that parameter defines which one i'm returning.

PR #117, opened 2020-08-10
Register user answer (not merge yet)

aroquev00:
In this PR the User's answer is registered in Firestore.
It checks if whether or not the user had already answered the question, and if not, then it registers the answer in the student's record in the "students" collection of the GameInstance and in the Question stats of the GameInstance.


src/main/java/com/google/sps/servlets/AnswerQuestionServlet.java
```diff
@@ -0,0 +1,156 @@
...
+
+    return false;
+  }
+
+  // Right now it only takes one single correct answer
+  private String getCorrectAnswer(String questionId, DocumentReference questionDocRef) {
+
+    String correctAnswerId = null;
+
+    ApiFuture<QuerySnapshot> future = questionDocRef.collection("answers").whereEqualTo("correct", true).get();
```
pamelalozano16, line 123: All db querys or actions should be done from the daos, for example in here a getCorrectAnswer in game

src/main/java/com/google/sps/servlets/AnswerQuestionServlet.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    DocumentReference questionDocRef = firestoreDb.collection("games").document(gameId).collection("questions")
+        .document(questionId);
+
+    String correctAnswerId = getCorrectAnswer(questionId, questionDocRef);
+
+    if (answerId.equals(correctAnswerId)) {
+      System.out.println("The answer is correct");
+
```
pamelalozano16, line 60: I'm gonna add the json object return here :)

aroquev00: Okay!

aroquev00:
Now when loading the panel we now check first for queryParams and if they're not present, only then we query the activeGameInstace from Firestore. I can definitely feel an improvement in speed guys!
image

PR #123, opened 2020-08-12
Game form submit

thefredo1000:
## About
After reading a lot of documentation, doing a lot of trial and error, a thousand cups of coffee, debugging and one long week of trying to create a dynamic form I present to you the Create Game page. (Don't try this at home)

Some things might seem to complex but that's because doing it in a simpler way didn't work.
### Before
Screenshot 2020-08-06 at 6 53 37 PM - Display 3
Finally we were able to create infinite questions for the form, but we weren't able to submit these questions to a form. I structured the questions at the form so it can be retrieved easily from JavaScript.
### After
Screenshot 2020-08-11 at 7 48 59 PM - Display 3
Now we can send all these questions with a button (which I would like recommendations on how to put it in a nicer way).
### Before
Screenshot 2020-08-11 at 8 14 43 PM - Display 1
We couldn't create a new document on the Firestore with the dynamic form from the front-end, also previously when we created a new game we had the questions with only 2 fields, the question title and an array of the answers.
### After
Screenshot 2020-08-11 at 7 48 59 PM - Display 1
When we click _Send Questions_ the questions are sent as a JSON object to the **New Game Servlet**, then the servlet sends it to the **Database Game DAO**. Now when we create a new game every question has a collection of answers with the same values as we had on the array, and some new fields such as:

- `isMC` : this defines if the question is multiple choice
- `title` : the title of the question
- `previousQuestion` : the id from the question that goes before the current one
- `nextQuestion` : the id from the question that goes after the current one

If there is no `nextQuestion` or a `previousQuestion`, it should be a `null` string.

## Extra
Sorry for the big PR. Also, feel free to tell me what can I do better: :)


aroquev00:
I think this is just awesome.

src/main/java/com/google/sps/data/Game.java
```diff
@@ -28,6 +34,52 @@ public static Builder builder() {
     return gameData;
   }
 
+  public static Game buildWithJson(String gameJsonStr) {
+    // The function recieves the JSON and converts to a JsonObject
+    String gameJson = gameJsonStr;
```
seanjohnite, line 39: nit: is this used? do you want to just use one of these?

src/main/webapp/teacher/createGame.html
```diff
@@ -88,24 +96,32 @@ <h3>Multiple Choice</h3>
                 <div class="mdl-cell mdl-cell--5-col">
                   <!-- First answer -->
                   <div class="mdl-textfield mdl-js-textfield mdl-textfield--floating-label">
+<<<<<<< HEAD
+                    <input class="mdl-textfield__input" type="text" id="game-correct-answer" name="answer0"> <label class="mdl-textfield__label" >Answer 1</label>
```
seanjohnite, line 100: Fix merge conflicts?

src/main/java/com/google/sps/data/Game.java
thefredo1000: Yes, this is used but I wasn't sure if I wanted to add this to `NewGameServlet.java` or to `Game.java` but I stuck with this, I think I'll change it to the other file.

src/main/webapp/teacher/createGame.html
thefredo1000: Done!

PR #125, opened 2020-08-12
Student answer feedback UI (With gifs)

pamelalozano16:
## PR Type

- [ ] Bugfix
- [x] Feature

## What is the new behavior?

The student selects the answer and when he submits his answer, it is registered in the backend. Then he has to wait until the teacher decides to end the question to see his results.

## Other information
Screenshot 2020-08-12 at 17 11 17 - Display 2
Screenshot 2020-08-14 at 13 59 54 - Display 2
Screenshot 2020-08-14 at 13 59 38 - Display 2


src/main/java/com/google/sps/servlets/AnswerQuestionServlet.java
```diff
@@ -35,7 +35,23 @@
...
+    if(gameInstance==null || gameInstance.isEmpty()){
+        response.setStatus(500);
+        response.getWriter().println("Room not specified");
+        return;
+    }
+
+    String token = request.getParameter("student");
+    if(token==null || token.isEmpty()){
+        response.setStatus(500);
+        response.getWriter().println("Room not specified");
```
aroquev00, line 48: not that it matters a lot, but the response should say something like: user not logged in right? Because the ID token would be empty?

src/main/java/com/google/sps/servlets/AnswerQuestionServlet.java
```diff
@@ -76,10 +92,16 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     DocumentReference questionDocRef = firestoreDb.collection("games").document(gameId).collection("questions")
         .document(questionId);
 
+    boolean isAnswerCorrect;
 
-    String correctAnswerId = getCorrectAnswer(questionId, questionDocRef);
-
-    boolean isAnswerCorrect = answerId.equals(correctAnswerId);
+    //If there was no answer it's wrong
+    if(answerId == null || answerId.isEmpty()){
```
aroquev00, line 98: In which cases can answerId be null and in which case can it be an empty string?

src/main/java/com/google/sps/servlets/AnswerQuestionServlet.java
```diff
@@ -98,10 +120,11 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
 
       // register the the question's general answer statistics
       updateGeneralQuestionStats(questionInGameInstanceDocRef, isAnswerCorrect, firestoreDb);
-
+      if(answerId != null && !answerId.isEmpty()){
```
aroquev00, line 123: I'm thinking that maybe we should make sure that an answerId cannot be empty when it is sent to the servlet? Or if it is possible, don't count it at all. Because if it is empty, we are already registering that the student got it wrong in his record (call in line 109), but we are not saving this "question answer" in the general statistics for the question, so this could lead to data inconsistency in the DB.

src/main/webapp/student/scripts/play-game.js
```diff
@@ -58,10 +61,20 @@ function getGameInstanceIdFromQueryParams() {
...
+  isCorrect = null;
+  document.getElementById("gif").setAttribute('src','')
+}
+
 // Inits listener to User's points in Firestore DB
-function initPointsListener(gameInstanceId, user) {
-  db.collection("gameInstance").doc(gameInstanceId).collection("students").doc(user.uid)
-  .onSnapshot(function(doc) {
+function updatePoints() {
+  db.collection("gameInstance").doc(gameInstanceId).collection("students").doc(studentId.uid).get().then(function(doc) {
```
aroquev00, line 77: Well thought for the points! Nice that now its queried when the question is finished and not before.

src/main/webapp/student/scripts/play-game.js
```diff
@@ -147,16 +162,34 @@ function initSubmitButton(gameInstanceId, gameId) {
...
 
   // Add the question title to the UI
   createQuestionObject(currentQuestion.title);
 
+  if(!currentQuestionActive){
+      if(!submited){
+        document.getElementById('submitButton').disabled = false;
+        document.getElementById("submitButton").click();
+        document.getElementById('submitButton').disabled = true;
+        document.getElementById("quiz").innerHTML='';
```
aroquev00, line 184: What is this for? An empty answer is being sent right?

src/main/java/com/google/sps/servlets/AnswerQuestionServlet.java
aroquev00: Oh I think I know why, if the user runs out of time then the answer would be an empty string right?

src/main/java/com/google/sps/servlets/AnswerQuestionServlet.java
aroquev00, line 123: Well thought for empty answers!

src/main/java/com/google/sps/servlets/AnswerQuestionServlet.java
pamelalozano16: You are right thank you!

src/main/webapp/student/scripts/play-game.js
pamelalozano16: Yeah so this means the current question is not active and there wasn't a submission or selection so it is submitted automatically, for example, in case that the time is up and the teacher ends the question, if the student didn't submit an answer it submits null which returns a wrong answer.

src/main/java/com/google/sps/servlets/AnswerQuestionServlet.java
pamelalozano16: Exactly

src/main/webapp/student/play-game.html
```diff
@@ -54,6 +54,10 @@
   <link rel="stylesheet" href="../styles/main.css">
   <link rel="stylesheet" href="styles/play-game.css">
 
+  <script
+  src="https://code.jquery.com/jquery-3.5.1.js"
+  integrity="sha256-QWo7LDvxbWT2tbbQ97B53yJnYU3WhH/C8ycbRAkjPDc="
+  crossorigin="anonymous"></script>
```
thefredo1000, line 60: Why did you add JQuery?

src/main/webapp/student/scripts/play-game.js
```diff
@@ -29,6 +35,7 @@ function initAuthStateObserver() {
 function authStateObserver(user) {
   if (user) { // User is signed in!
     // Everything starts working when the User logs in
+    studentId = user;
     loadGamePanel(user)
   } else { // User is signed out!
     console.log("Not logged in");
```
thefredo1000, line 41: Should we add something to handle this?

thefredo1000: Or maybe it not necessary

src/main/webapp/student/scripts/play-game.js
```diff
@@ -242,5 +278,47 @@ function createAnswer(quiz, multipleDiv, doc, i){
...
+        const info = await infoJson.json(); 
+        isCorrect = info.correct;
+        resultObject.classList.toggle(isCorrect.toString());
+        if(isCorrect) {
+            tag = "celebrate";
+            resultObject.innerText = 'Correct!';
+        } else {
+            tag = "disappointment";
+            resultObject.innerText = 'Incorrect :(';
+        }
```
thefredo1000, line 297: Nit: We can add the word "Disney" to the tag, so even if it has a hard time filtering some stuff, we know that it's really family friendly

src/main/webapp/student/play-game.html
pamelalozano16: Oh I was going to use it but let me delete it, thanks!

src/main/webapp/student/scripts/play-game.js
pamelalozano16: Right

PR #126, opened 2020-08-14
Teacher history stats

aroquev00:
The teacher UI now displays all the necessary info for each of the questions.

How it works? 
First when the user is logged in the JS builds the "active question" panel, which displays all information about the current question. When this is loaded, the JS then proceeds to build the question history. 

The question history queries Firestore once and adds all the details about EVERY question on the game, if the question has already passed, the results are shown, and if not, the answers indicators are all zeros.

How does it get updated?
When the teacher changes the activeQuestion in the panel, a listener to Firestore is initiated and when someone answers the question, it updates this info in the "main panel". Well, now that same listener updates that question in the "main panel" AND in the history section.

If code is too messy please let me know so that I can refactor.

image

Next I will work on displaying individual student's stats


src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
```diff
@@ -188,22 +203,131 @@ async function updateCurrentQuestion({ gameId, currentQuestionId, isCurrentQuest
...
+function queryCurrentQuestion({ gameId, currentQuestionId }) {
+  return db.collection('games').doc(gameId).collection('questions').doc(currentQuestionId).get().then(function(doc) {
+    if (doc.exists) {
+      return doc.data()
+    }
+  });
+}
+
+// Init the listener to a change in the current question's answers
+function initQuestionAnswerStatsListener({ gameInstanceId, currentQuestionId } = {}) {
```
thefredo1000, line 216: What does `{ gameInstanceId, currentQuestionId } = {}` mean?

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
```diff
@@ -188,22 +203,131 @@ async function updateCurrentQuestion({ gameId, currentQuestionId, isCurrentQuest
...
+  if (unsubscribeCurrentActiveQuestionAnswersInGameInstance != null) {
```
thefredo1000, line 217: Nit: you can just put `if (unsubscribeCurrentActiveQuestionAnswersInGameInstance)  { ... }`

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
```diff
@@ -188,22 +203,131 @@ async function updateCurrentQuestion({ gameId, currentQuestionId, isCurrentQuest
...
+  if (answerInQuestionStatsDivElement == null) {
+    // This may be executed before the node is created the first time
+    return;
+  }
+
+  // Update question stats
+  answerInQuestionStatsDivElement.innerText = updatedAnswer.title + ' with ' + updatedAnswer.numberAnswers + ' answers.';
 
+  if (updatedAnswer.correct) {
+    answerInQuestionStatsDivElement.innerText += ' (Correct answer)';
```
thefredo1000, line 268: Nit: Maybe a (wrong answer)?

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
brettapeters: This is a common way to have named parameters with defaults in JS. Calling this function with an argument `{ gameInstanceId: 123, currentQuestionId: 5 }` will result in the argument object getting destructured and its values will be assigned to the identifiers `gameInstanceId` and `currentQuestionId` accordingly.

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
aroquev00: Nice! done.

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
aroquev00: True, done!

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
```diff
@@ -188,22 +203,133 @@ async function updateCurrentQuestion({ gameId, currentQuestionId, isCurrentQuest
...
 function updateNumberOfMembersUI(numberOfMembers) {
   const numberOfMembersElement = document.getElementById("jsNumberOfStudents");
   numberOfMembersElement.innerText = "There are " + numberOfMembers + " students registered in your room.";
 }
 
+// Listen to the current question's stats for any change
+function initQuestionStatsListener({ gameInstanceId, currentQuestionId } = {}) {
+  if (unsubscribeCurrentActiveQuestionInGameInstance != null) {
+    // This is to stop listening to live changes to the previous question which is not active anymore
+    unsubscribeCurrentActiveQuestionInGameInstance();
```
seanjohnite, line 284: What does calling this do? Is it a function?

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
```diff
@@ -188,22 +203,133 @@ async function updateCurrentQuestion({ gameId, currentQuestionId, isCurrentQuest
...
+
+  const numberWrongQuestionAnswersHistoryElement = document.getElementById('stats-' + currentQuestionId + '-numberWrongQuestionAnswers');
+  numberWrongQuestionAnswersHistoryElement.innerText = 'Number of wrong anwers: ' + updatedQuestionStats.numberWrong;
+}
+
+// Show the updated number of members in UI
+function updateNumberOfMembersUI(numberOfMembers) {
+  const numberOfMembersElement = document.getElementById("jsNumberOfStudents");
+  numberOfMembersElement.innerText = "There are " + numberOfMembers + " students registered in your room.";
+}
```
seanjohnite, line 331: duplicated function updateNumberOfMembersUI?

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
```diff
@@ -235,4 +361,87 @@ function initUIControlButtons(gameInstanceId) {
   });
 }
 
+// Probably should do this following the Linked List instead of querySnapshot.forEach
```
seanjohnite, line 364: If you're getting all of the questions, getting the array is probably better : )

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
```diff
@@ -235,4 +361,87 @@ function initUIControlButtons(gameInstanceId) {
...
+  singleQuestionStatDivElement.id = 'stats-' + questionId
+  singleQuestionStatDivElement.classList.add('questionInHistory');
+
+  const questionTitle = document.createElement('div');
+  questionTitle.innerText = 'Question title: ' + question.title;
+  singleQuestionStatDivElement.appendChild(questionTitle);
+
+  // How many students have answered the question
+  const numberQuestionAnswersHistoryElement = document.createElement('div');
+  numberQuestionAnswersHistoryElement.id = 'stats-' + questionId + '-numberQuestionAnswers';
```
seanjohnite, line 397: What are these ids used for? Later retrieval?

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
```diff
@@ -273,6 +288,8 @@ function initQuestionStatsListener({ gameInstanceId, currentQuestionId } = {}) {
 }
 
 function updateCurrentQuestionStats(updatedQuestionStats) {
```
pamelalozano16, line 296: Maybe you can do an if to verify that the change in the doc was a change in the stats? That way this function is not activated unless it's necessary

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
aroquev00: Yep this is to stop listening to the DocRef of the previous question (supposing you clicked next question), it wouldn't make any difference if we didn't unsubscribe because once we're past a question it can't be updated anymore, but I thought it would still be good to unsubscribe to save some space in memory maybe?

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
aroquev00: Yep, done!

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
```diff
@@ -235,13 +237,63 @@ function initGameInstanceListener(gameInstanceId) {
...
   });
 }
 
-// Updates the panel showing which questions students are seeing
+function initQuestionAnswerStatsListener(gameInstanceId, currentQuestionId) {
+
+  if (unsubscribeCurrentActiveQuestionAnswersInGameInstance != null) {
+    unsubscribeCurrentActiveQuestionAnswersInGameInstance();
+  }
+  unsubscribeCurrentActiveQuestionAnswersInGameInstance = db.collection('gameInstance').doc(gameInstanceId).collection('questions').doc(currentQuestionId).collection('answers')
```
pamelalozano16, line 221: I think you don't need two listeners? If answers is a subcollection on the same currentQuestion document you can just listen to that document, have a global variable for the number of answers (that you retrieve when the page loads), and then just make a single query to check if it's different from the one you have stored, if it is then you updateCurrentQuestionAnswerStats(). I think this would be more efficient than having two listeners

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
aroquev00: Gotcha, was thinking about the fact that Firestore stores documents by ordering alphabetically so _maybe_ the array could contain the questions not in the original order? Actually I think this only would happen if we modify the game question's order after we have created it, because at the moment of game creation, Firestore creates documents for each question also in alphabetical order (I think), so maybe that's why we haven't got any issues with the order of question display in the teacher UI...

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
```diff
@@ -299,8 +351,21 @@ function updateCurrentQuestionStats(updatedQuestionStats) {
...
+  if (numberQuestionAnswersHistoryElement == null) {
+    // This may be executed before the node is created the first time
+    return;
+  }
+  numberQuestionAnswersHistoryElement.innerText = 'Number of answers: ' + updatedQuestionStats.numberAnswered;
+
+  const numberCorrectQuestionAnswersHistoryElement = document.getElementById('stats-' + currentQuestionId + '-numberCorrectQuestionAnswers');
+  numberCorrectQuestionAnswersHistoryElement.innerText = 'Number of correct anwers: ' + updatedQuestionStats.numberCorrect;
 
+  const numberWrongQuestionAnswersHistoryElement = document.getElementById('stats-' + currentQuestionId + '-numberWrongQuestionAnswers');
```
pamelalozano16, line 323: NIT: If you are creating the element here (line 117) mayyyybe it's better if you pass the element as an argument instead of creating these complex ids?

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
aroquev00: Yep, in line 320 I update this DOM element and access it with its ID. All of this is so that I only build the question history "tree" once, and when that question becomes active and is updated as students answer, I just go and update that number in the history section as well, instead of having to recreate the whole question history object. Do you think it's OK or too complicated?

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
aroquev00: Right well thought! I will store the "state" in some general variables to only update the necessary parts when we receive updates.

aroquev00: Done!

aroquev00:
Okay team so I added a helper function in line 302 to make the whole updating a question's stats process a bit more understandable. If you think anything else should be done make sure to tell me! If not, then I believe this PR is ready to merge @seanjohnite

PR #128, opened 2020-08-17
Register more in answer servlet

aroquev00:
This PR modifies some files to correctly update information in Firestore when a student answers a question. Now the records show accurate data.

Also, we now initiate a student in a gameInstance with the necessary values in 0 to participate in the game.

### DIFF
--- a/src/main/java/com/google/sps/daos/DatabaseUserDao.java
+++ b/src/main/java/com/google/sps/daos/DatabaseUserDao.java
@@ -108,6 +108,9 @@ public class DatabaseUserDao implements UserDao {
   private void registerUserInGameInstance(DocumentReference userInGameInstanceDocRef, String userId) {
     Map<String, Object> docData = new HashMap<>();
     docData.put("points", 0);
+    docData.put("numberAnswered", 0);
+    docData.put("numberCorrect", 0);
+    docData.put("numberWrong", 0);
 
     userInGameInstanceDocRef.set(docData);
   }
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/sps/daos/DatabaseUserDao.java
```diff
@@ -108,6 +108,9 @@ private void initUser(String userId) {
   private void registerUserInGameInstance(DocumentReference userInGameInstanceDocRef, String userId) {
     Map<String, Object> docData = new HashMap<>();
     docData.put("points", 0);
+    docData.put("numberAnswered", 0);
+    docData.put("numberCorrect", 0);
+    docData.put("numberWrong", 0);
```
thefredo1000, line 113: You can have a function on **`GameInstance.java`** that generates this automatically, so you'll only need to call the function from an Game Instance object, which in this case it will be empty and you may not need it, but in other cases it's really helpful.
### **Example from the `Game.java` class**
```Java
  public Map<String, Object> gameData() {

    Map<String, Object> gameData = new HashMap<>();
        gameData.put("creator", this.creator());
        gameData.put("title", this.title());
        gameData.put("headQuestion", this.headQuestion());
        gameData.put("numberOfQuestions", this.questions().size());
    return gameData;
  }
```

aroquev00: Mmm nice, I think we first need to do some work with `GameInstance.java` first to have all the info we need there? But this is something that definitely will come in handy when we refactor and abstract logic!

seanjohnite: +1 nice comment rodrigo!

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
```diff
@@ -219,18 +219,26 @@ function initQuestionAnswerStatsListener({ gameInstanceId, currentQuestionId } =
...
+    querySnapshot.docChanges().forEach(function(change) {
+      if (change.type === "added") {
+        // Add the answer to the active panel UI
+        addCurrentQuestionAnswerStats({ answer: change.doc.data(), answerId: change.doc.id, questionId: currentQuestionId });
+      }
 
-    querySnapshot.forEach(function(doc) {
-      // Update the current question's answers in the "Active" top panel
-      updateCurrentQuestionAnswerStats(doc.data(), doc.id);
+      if (change.type === "modified") {
```
pamelalozano16, line 234: This is greattt

PR #133, opened 2020-08-19
Add User servlet test

aroquev00:
I just added a test for the `UserServlet.java`, which registers a user in Firestore if the user is not already registered.


src/test/java/com/google/sps/UserServletTest.java
```diff
@@ -0,0 +1,67 @@
...
+        return mockServletContext;
+      }
+    };
+
+    when(mockServletContext.getAttribute("userDao")).thenReturn(mockUserDao);
+  }
+
+  @Test
+  public void doPostUser() throws IOException {
+    String json = "{\"idToken\":\"asdfqwerty\"}";
```
pamelalozano16, line 51: I think it's easier to test with parameters instead of json, maybe something to consider later

PR #138, opened 2020-08-21
Init game instance all data

aroquev00:
Now when we want to start a new Game Instance we can do it and the necessary structure is created in Firestore, ready to play :)
I had to "manually" update some data in Firestore (instead of using `.set(object)`), and probably will have to create some classes to abstract more logic. But hey, it works! :P


src/main/java/com/google/sps/daos/DatabaseGameInstanceDao.java
```diff
@@ -38,13 +42,51 @@ public String createNewGameInstance(String idToken, String gameId) {
...
+    // newGameInstance.setId(newGameInstanceRef.getId());
+
+    Map<String, Object> gameInstanceData = new HashMap<String, Object>();
+    gameInstanceData.put("creator", userId);
+    gameInstanceData.put("currentQuestion", null);
+    gameInstanceData.put("currentQuestionActive", false);
+    gameInstanceData.put("gameId", gameId);
+    gameInstanceData.put("id", newGameInstanceRef.getId());
+    gameInstanceData.put("isActive", false);
+    gameInstanceData.put("numberOfMembers", 0);
```
pamelalozano16, line 60: Here we can update with a game instance object I think

PR #139, opened 2020-08-22
Animal alias

### LINKED ISSUE #130: Add animals nicknames to students
Add an animal nickname to each student so that they are identifiable.

aroquev00:
The students now have animal aliases! Should help keeping people identifiable. Also, the alias is assigned after a transaction so it is _guaranteed_ that no two students that join at the same time have the same animal, only if there are more students than aliases then the names are assigned to multiple people.

Student view
image

Teacher view
image

Closes #130


src/main/java/com/google/sps/daos/DatabaseUserDao.java
```diff
@@ -105,25 +126,38 @@ private void initUser(String userId) {
...
+    //   long oldNumberOfMembers = snapshot.getLong("numberOfMembers");
+    //   transaction.update(gameInstanceDocRef, "numberOfMembers", oldNumberOfMembers + 1);
+    //   return null;
+    // });
   }
-
   
+  private String getAnimal(int index) {
+    
+    return animalList.get((index % animalListSize));
```
pamelalozano16, line 160: Why do you do this operation ? Can't you just return `animalList.get(index)` ?

PR #144, opened 2020-08-26
Jest Testing on createGameElements.js

thefredo1000:
## About
I wanted to test on `createGame.js` but I couldn't do it because of the Firebase Auth lines that script has, so what I did was to separate the important element building functions from `createGame.js` into `createGameElements.js`. Actually that was what I wanted to do in the beginning. So here are some small test, tell me what do you think.

### Is the true or false grid appearing
This just checks if the grid for the true or false answers is appearing.

### Is the multiple choice grid appearing
This just checks if the grid for the multiple choice answers is appearing.

### Is the title grid appearing
This just checks if the grid for the question title is appearing.

### Randomly generated question type
This generates 50 questions and each question is randomly set to a True or False question or Multiple Choice then it checks in the HTML if they were generated correctly.


src/main/webapp/__tests__/createGameElements.test.js
```diff
@@ -0,0 +1,128 @@
...
+        '</div>' +
+      '</div>' +
+      '<div class="mdl-cell mdl-cell--3-col"></div>' +
+    '</div>' + '</div>';
+
+    const expectedTitleDiv = titleGrid(title);
+    expect(document.getElementById('titleGrid').innerHTML).toContain(expectedTitleDiv.innerHTML);
+  });
+
+  test("Randomly generated question type", () => {
```
brettapeters, line 108: I think the 50 random question set up is unnecessary. Just test properties on the output of `questionElement(0, 'q1', true)`, and `questionElement(0, 'q1', false)`

src/main/webapp/__tests__/createGameElements.test.js
```diff
@@ -0,0 +1,128 @@
+// Provides more matchers like `toBeOneOf`
+require('jest-extended'); 
+// module path is relative to 'src/main/webapp/'
+const { trueOrFalseGrid, multipleChoiceGrid, titleGrid, questionElement } = require('scripts/createGameElements.js');
+
+// `describe` blocks can be used to group related tests together.
+describe('createGameGrids', () => {
+  // `test` is a single unit test
+  test('Is the true or false grid appearing', () => {
```
brettapeters, line 9: Be wary of writing tests that closely mirror the code they are testing. These tests will break if there are any changes to the HTML output by the functions. 

A more reliable test might do this:
1) append `trueOrFalseGrid(0)` to `document.body`
2) query for list of checkbox inputs.
3) `expect` that the checkbox inputs have the right ID attributes ("question0checkbox0" and "question0checkbox1").

brettapeters (approved):

PR #145, opened 2020-08-27
Join link

aroquev00:
LAST PRRRRR
Now when the teacher creates a gameInstance there is a link the teacher can share with the students for them to join. If the teacher clicks the link it gets copied to the clipboard so that its ready to be shared with the students.

I moved the process so that when the student introduces the room code in the join page, it only redirects to the play screen, and now that is where the POST request is sent to the back end for the student to be registered and assigned an animal alias. Actually this is better because we now have one query instead of two as previously.

image

### DIFF
--- a/src/main/java/com/google/sps/daos/DatabaseUserDao.java
+++ b/src/main/java/com/google/sps/daos/DatabaseUserDao.java
@@ -117,6 +117,8 @@ public class DatabaseUserDao implements UserDao {
     
   }
 
+  // Register a user in a gameInstance and return its newly assigned animal alias
+  // If user is already registered, just return the animal alias that was already assigned before
   @Override
   public String joinGameInstance(String idToken, String gameInstanceId) {
     FirebaseToken decodedToken = null;
@@ -158,6 +160,8 @@ public class DatabaseUserDao implements UserDao {
 
     // Add the active GameInstance to the User's entry in "Users" collection
     firestoreDb.collection("users").document(userId).update("activeGameInstanceId", gameInstanceId);
+
+    // Return the animal that was assigned to the user
     return animal;
   }
 
--- a/src/main/java/com/google/sps/daos/DatabaseUserDao.java
+++ b/src/main/java/com/google/sps/daos/DatabaseUserDao.java
@@ -11,6 +11,7 @@ import com.google.api.core.ApiFuture;
 import com.google.cloud.firestore.DocumentReference;
 import com.google.cloud.firestore.DocumentSnapshot;
 import com.google.cloud.firestore.Firestore;
+import com.google.cloud.firestore.WriteResult;
 import com.google.firebase.auth.FirebaseAuth;
 import com.google.firebase.auth.FirebaseAuthException;
 import com.google.firebase.auth.FirebaseToken;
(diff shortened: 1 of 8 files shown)


src/main/java/com/google/sps/daos/DatabaseUserDao.java
```diff
@@ -117,7 +118,7 @@ public void createIfNotExists(String idToken) {
   }
 
   @Override
-  public void joinGameInstance(String idToken, String gameInstanceId) {
+  public String joinGameInstance(String idToken, String gameInstanceId) {
```
seanjohnite, line 121: Document this return value.

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
```diff
@@ -148,6 +148,12 @@ function buildActiveGameInstanceUI({ gameInstanceId, gameInstance, game} = {}) {
 function addGameInstanceIdToUI(gameInstanceId) {
   const gameInstanceIdElement = document.getElementById("jsGameInstanceId");
   gameInstanceIdElement.innerText = "This gameInstance's ID is: " + gameInstanceId;
+
+  const shareGameInstanceElement = document.getElementById('jsShareGameInstance');
+  shareGameInstanceElement.innerText = "Share this link with your students for them to join (click to copy to clipboard): " + "https://quizzy-step-2020.uc.r.appspot.com/student/play-game.html?gameInstanceId=" + gameInstanceId;
```
seanjohnite, line 153: does this element have have cursor: pointer set on CSS?

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
```diff
@@ -479,9 +485,7 @@ function addQuestionAnswerToHistoryUI({ questionId, answerId, answer } = {}) {
   answerInQuestionStatsDivElement.innerText = answer.title + ' with ' + answer.numberAnswers + ' answers.';
 
   if (answer.correct) {
-    questionAnswerDivElement.innerText = 'Correctly answered, chose: "' + answer.chosen + '"';
-  } else {
-    questionAnswerDivElement.innerText = 'Incorrect answer, chose: "' + answer.chosen + '"';
```
seanjohnite, line 484: removed this on purpose?

seanjohnite (approved):
LG!

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
aroquev00: Yep it does!

src/main/webapp/teacher/controlGameInstance/controlGameInstance.js
aroquev00: Yep that was there by mistake, was accessing nonexistent info and changed it to display the actual answer info.

src/main/java/com/google/sps/daos/DatabaseUserDao.java
aroquev00: Done! Let me know if its missing more explanation.

